PR #88, opened 2024-12-04
Improve HiddenSet ergonomics and fix a bug in the leveled compaction strategy

carlsverre:
The strategy failed to consider hidden segments when computing the minimal compaction job.

### DIFF
--- a/src/compaction/leveled.rs
+++ b/src/compaction/leveled.rs
@@ -6,7 +6,7 @@ use super::{Choice, CompactionStrategy, Input as CompactionInput};
 use crate::{
     config::Config,
     key_range::KeyRange,
-    level_manifest::{level::Level, LevelManifest},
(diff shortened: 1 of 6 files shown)


carlsverre:
Should be merged into https://github.com/fjall-rs/lsm-tree/pull/88

src/compaction/leveled.rs
```diff
@@ -20,21 +20,41 @@ fn aggregate_key_range(segments: &[Segment]) -> KeyRange {
 fn pick_minimal_compaction(
     curr_level: &Level,
     next_level: &Level,
-    hidden_set: &HiddenSet,
+    levels: &LevelManifest,
```
marvin-j97: This can probably stay `&HiddenSet`, and just use `HiddenSet::contains` instead

carlsverre: kk

marvin-j97 (approved):

PR #103, opened 2025-01-18
feat: Prefix Bloom Filter

### LINKED ISSUE #97: Prefix bloom filters
May not be necessary when there are range filters though...

khlopkov:
This PR supposed to resolve #97 

Please, let me know if I missed something.

### DIFF
--- a/src/prefix_extractor.rs
+++ b/src/prefix_extractor.rs
@@ -1,0 +1,11 @@
+/// A trait allowing to customize prefix extraction in operations with bloom filter.
+/// It defines how prefix should be extracted from the key, when update or read
+/// a bloom filter.
+pub trait PrefixExtractor: Sync + Send {
+    /// Extracts prefix from original key
+    fn transform<'a>(&self, key: &'a [u8]) -> &'a [u8];
+    /// Checks if a key is in domain and prefix can be extracted from it.
+    /// For example if `PrefixExtractor` suppose to extract first 4 bytes from key,
+    /// `in_domain(&[0, 2, 3])` should return false
+    fn in_domain(&self, key: &[u8]) -> bool;
+}
--- a/src/segment/writer/mod.rs
+++ b/src/segment/writer/mod.rs
@@ -462,7 +462,7 @@ mod tests {
             data_block_size: 4_096,
             index_block_size: 4_096,
             segment_id,
-            prefix_extractor: Option::None,
+            prefix_extractor: None,
         })?
         .use_bloom_policy(BloomConstructionPolicy::BitsPerKey(0));
 
--- a/src/segment/writer/mod.rs
+++ b/src/segment/writer/mod.rs
@@ -15,6 +15,7 @@ use super::{
 use crate::{
     coding::Encode,
     file::fsync_directory,
+    prefix_extractor::PrefixExtractor,
     segment::{block::ItemSize, value_block::BlockOffset},
     value::{InternalValue, UserKey},
     SegmentId,
@@ -23,6 +24,7 @@ use std::{
     fs::File,
     io::{BufWriter, Seek, Write},
     path::PathBuf,
+    sync::Arc,
 };
 
 #[cfg(feature = "bloom")]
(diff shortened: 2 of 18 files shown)


src/prefix_extractor.rs
```diff
@@ -0,0 +1,11 @@
+/// A trait allowing to customize prefix extraction in operations with bloom filter.
+/// It defines how prefix should be extracted from the key, when update or read
+/// a bloom filter.
+pub trait PrefixExtractor: Sync + Send {
+    /// Extracts prefix from original key
+    fn transform<'a>(&self, key: &'a [u8]) -> &'a [u8];
+    /// Checks if a key is in domain and prefix can be extracted from it.
+    /// For example if `PrefixExtractor` suppose to extract first 4 bytes from key,
+    /// `in_domain(&[0, 2, 3])` should return false
+    fn in_domain(&self, key: &[u8]) -> bool;
```
marvin-j97, line 10: I don't think this is necessary.
If a key is [1, 2, 3] and we extract the first 4 bytes, we would just take the entire string, so:

```js
let len = Math.min(key.len(), 4);
return key.slice(0, len);
```

khlopkov: Let me better explain why is this needed in more practical example:

Imagine we have a database where keys are constructed using the following template: `{datatype}#{id}` and a prefix extractor that would extract `{datatype}#` prefix from a key to optimize scans of all rows of `datatype#`. Then, in table were populated following rows `table1#a`, `table1#b`, `table2#a`. The keys for bloom filter would be `table1#`, `table2#` in this case. Then, if we use prefix scan by prefixes `table1#` or `table2#` everything will work correctly. But if scan is performed with `table` prefix without `in_domain` the full prefix will be extracted and will try to check bloom filter by `table` key, which may return false negative. `in_domain` in this case will help to prevent such cases and will use Bloom Filter only in case when prefix is in domain. In this example everything starting with `{datatype}#` will be considered as `in_domain` and use Bloom Filter, other prefixes will be `not in_domain` and will just use normal prefix search ignoring Bloom Filter

marvin-j97: Makes sense.

However, for the prefix extractor to support structured prefixes, the trait needs to return an Iterator over extracted prefixed (see https://github.com/fjall-rs/lsm-tree/issues/97 and https://github.com/fjall-rs/fjall/issues/116). At that point, the trait could just return `std::iter::empty`, which would signal that the key is "not in domain".

For example we may extract prefixes, like:

`eu#germany#berlin`

-> extract "eu#" as prefix
  -> also extract "eu#germany#"

Now prefix searches over both continent and [continent + country] can be filtered by the prefix filter.

src/segment/writer/mod.rs
```diff
@@ -248,8 +251,20 @@ impl Writer {
             // of the same key
             #[cfg(feature = "bloom")]
             if self.bloom_policy.is_active() {
-                self.bloom_hash_buffer
-                    .push(BloomFilter::get_hash(&item.key.user_key));
+                let key: Option<&[u8]> = match &self.opts.prefix_extractor {
```
marvin-j97, line 254: To control the bits per key independently for prefix and point reads, the prefix bloom filter needs to be built from its own buffer, and stored separately in the disk segment.
That's why there's a dedicated pointer for prefix filters in the segment metadata which is currently unused and always 0 (NULL): https://github.com/fjall-rs/lsm-tree/blob/77fb2a97f1707157f967f9dcbd047850aca3a5ca/src/segment/file_offsets.rs#L24

src/tree/mod.rs
```diff
@@ -602,7 +605,17 @@ impl Tree {
         // NOTE: Create key hash for hash sharing
         // https://fjall-rs.github.io/post/bloom-filter-hash-sharing/
         #[cfg(feature = "bloom")]
-        let key_hash = crate::bloom::BloomFilter::get_hash(key.as_ref());
```
marvin-j97, line 605: When the prefix bloom filter is separated, this can all be reverted because the point read path does not change at all.

marvin-j97:
Superseded by #151

PR #112, opened 2025-03-06
perf: replace std partition_point

marvin-j97:


marvin-j97:
Screenshot from 2025-03-19 22-48-41
Screenshot from 2025-03-19 22-48-54
10% -> 1.7% in hot blob point read path

marvin-j97:
Screenshot from 2025-03-20 01-24-14
Screenshot from 2025-03-20 01-24-35

23% -> 9% in hot non-blob point read path (partitioned index)

marvin-j97:
rustc regression: https://github.com/rust-lang/rust/issues/138796

Issue #95, opened 2024-12-14
Use skiplist tailored for LSM-tree


marvin-j97:
- No need for remove and get() methods, we only need range() and insert()

- Arena-based (store towers/nodes in arena)

guycipher:
Why not a read-write lock skiplist?  Why the granular concurrency?  It will lead to complexity and unsafely no?  From experience I can’t say you’ll get major performance boosts doing a granular approach.

marvin-j97:
> Why not a read-write lock skiplist?  Why the granular concurrency?  It will lead to complexity and unsafely no?  From experience I can’t say you’ll get major performance boosts doing a granular approach.  

Well the current skiplist already is concurrent.
Granular not being better is anecdotal, the point is to have a lock-free read path, and make reads not blocking writes too much.

> Also, why just range?  What about iteration for cursing through??  Forwards and backwards the entire list.  Range on a memtable? 

Range is an Iterator. More correctly a DoubleEndedIterator.

guycipher:
**Well the current skiplist already is concurrent.
Granular not being better is anecdotal, the point is to have a lock-free read path, and make reads not blocking writes too much.**
I'm just providing some perspective.  Yes that is nice but comes with complexity and dare I see it doesn't provide much of a boost as it does bugs and uncertainty; Since the skip list is already in memory it's rather performent no? 

**Range is an Iterator. More correctly a DoubleEndedIterator.**
I thought this was more specific to databases.  A range iterator is like hey give me 100-3000 and you can only iterate between those, that's what I understood by it.  I mean this can be done in any way truly.  Here's some more food for thought can iterating through the memtable and sstables from the memtable to the last sstable?

marvin-j97:
> Yes that is nice but comes with complexity and dare I say it doesn't provide much of a boost as it does bugs and uncertainty; Since the skip list is already in memory it's rather performent no?

It's not about boosting performance, it's about keeping the current locking behaviour, but not depending on crossbeam's epoch system. For long reads, having a RwLock would block writes.

> Here's some more food for thought can iterating through the memtable and sstables from the memtable to the last sstable be accomplished?

And yes, that's the point of an LSM-tree.

PR #131, opened 2025-05-18
memtable/skiplist: add a purpose-built skiplist

ajwerner:
Fixes https://github.com/fjall-rs/lsm-tree/issues/95.


marvin-j97:
With the `Borrow<Q>` trait bound it's not possible to use `InternalKeyRef` that would allow to not build an `InternalKey` (which involves a heap allocation).

src/memtable/skiplist/arena.rs
```diff
@@ -0,0 +1,118 @@
...
+        atomic::{AtomicPtr, AtomicUsize, Ordering},
+        Mutex,
+    },
+};
+
+// DEFAULT_BUFFER_SIZE needs to be at least big enough for one fullly-aligned node
+// for the crate to work correctly. Anything larger than that will work.
+//
+// TODO: Justify this size.
+const DEFAULT_BUFFER_SIZE: usize = (32 << 10) - size_of::<AtomicUsize>();
```
marvin-j97, line 18: Need to play with this a bit - but should probably be much higher by default: 1 MB or so?

src/memtable/skiplist/arena.rs
```diff
@@ -0,0 +1,118 @@
...
+impl<const BUFFER_SIZE: usize> Default for Arenas<BUFFER_SIZE> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+unsafe impl<const N: usize> Send for Arenas<N> {}
+unsafe impl<const N: usize> Sync for Arenas<N> {}
+
+pub(crate) struct Arenas<const BUFFER_SIZE: usize = DEFAULT_BUFFER_SIZE> {
```
marvin-j97, line 29: Eventually, for write transactions, the size should be much smaller (so that small transactions don't overallocate too much) - so this needs to be a non-generic parameter.

src/memtable/mod.rs
```diff
@@ -131,7 +134,11 @@ impl Memtable {
             .fetch_add(item_size, std::sync::atomic::Ordering::AcqRel);
 
         let key = InternalKey::new(item.key.user_key, item.key.seqno, item.key.value_type);
-        self.items.insert(key, item.value);
+        // TODO(ajwerner): Decide what we want to do here. The panic is sort of
+        // extreme, but also seems right given the invariants.
```
marvin-j97, line 133: Write transactions write with a fixed sequence number initially, so we actually need to overwrite values to keep the same behaviour as crossbeam's skiplist

https://github.com/fjall-rs/fjall/blob/5001d4db6430808df4c8ba6db12c2dbaaf7a91ec/src/tx/write/mod.rs#L379-L381

marvin-j97 (changes requested):

src/memtable/mod.rs
ajwerner: I see. This is doable if we’re okay leaking memory for each write. If we’re not then we’ll need some sort of free list structure. I’ll take a stab. I’ve been thinking I was to change the memory layout so drop doesn’t have to bounce around through iteration.

src/memtable/skiplist/arena.rs
ajwerner: Okay, I can do that.

src/memtable/skiplist/arena.rs
ajwerner: Yeah, it should be bigger than 32k, but 1MiB might be too big. The keys and values are not inline, it’s just the metadata. The questions I’d have are how expensive is allocating a new block, and how expensive is inserting into the skip map. My guess is that the alloc is not likely worse than 10us (it’s probably way less) and the inserts are ~100ns. If you can fit 1000 in here (if we say the average links is 32 and the key and value are each 32 bytes), then you’ll have spent at least 10x as long doing the inserting. In practice I think the mallocs even with zeroing is a lot cheaper. The benchmarks I was playing with don’t show much win above 256KiB.

ajwerner (commented):
I’ll play with updating. It’s pretty hard to not leak the node from the previous update without hooking up a free list but it’s also not so hard to add one.

marvin-j97:
Another consideration is small write transactions (which hold one or more skiplists as write set): we don't want to allocate a massive 100+ kib slab to put a few items into, so the scaling either needs to be smaller of exponential, such that we can adjust the scaling only in transactions.

src/memtable/mod.rs
marvin-j97: ~~FWIW the implementation of the write sets in the transactions *could* be separate from the memtables. That would allow something the like STL B-tree to be used (because transactions are not concurrent anyway), maybe~~

~~Then the whole free list stuff would not be necessary, but it would require some code changes somewhere else in the code base~~

The impl in the write set still needs to be lock-less I think.

marvin-j97: This can probably be solved by reserving the 64th (highest) bit in the sequence number for transactions.
Then any write in the transaction starts at 0x80...0. That way the transaction sequence numbers are always higher than the database's.
RocksDB actually only uses 56-bit sequence numbers.

marvin-j97:
@ajwerner Can you reopen this PR against `main`?

PR #145, opened 2025-08-08
Update xxhash usage to rely on default hasher

DoumanAsh:
This hasher has smaller size which is best to how std library uses hasher


marvin-j97:
Thanks, I have noticed I only use these maps for Segment IDs (u64), so I decided to use rustc_hash for those (sliiightly faster than xxh3).

PR #150, opened 2025-09-01
add prefix extraction support for bloom filters with configurable extractors

zaidoon1:
address https://github.com/fjall-rs/lsm-tree/issues/97

### DIFF
--- a/src/segment/mod.rs
+++ b/src/segment/mod.rs
@@ -22,6 +22,7 @@ use crate::{
     bloom::{BloomFilter, CompositeHash},
     cache::Cache,
     descriptor_table::FileDescriptorTable,
+    prefix::SharedPrefixExtractor,
     time::unix_timestamp,
     tree::inner::TreeId,
     value::{InternalValue, SeqNo, UserKey},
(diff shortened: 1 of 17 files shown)


marvin-j97:
Can you rebase this onto the 3.0.0 branch?

src/segment/mod.rs
```diff
@@ -325,8 +328,23 @@ impl Segment {
...
+                match bf.contains_prefix(key, extractor.as_ref()) {
+                    Some(false) => {
+                        // Bloom filter says key is definitely not present
+                        return Ok(None);
+                    }
+                    Some(true) | None => {
+                        // Either bloom filter says key might be present (Some(true))
+                        // or key is out of domain (None) - proceed with lookup
+                    }
+                }
```
marvin-j97, line 342: This could be an if-let:

```rs
if let Some(false) = bf.contains_prefix(key, extractor.as_ref()) {
   return Ok(None);
}
```

zaidoon1:
> Can you rebase this onto the 3.0.0 branch?

can do, going to close this merge request, it's easier to start fresh. Also I realized I missed the most important part i care about which is to check the prefix bloom during a seek

marvin-j97:
> Also I realized I missed the most important part i care about which is to check the prefix bloom during a seek so I'll add that too

The 3.0.0 branch has an optional `metrics` feature that allows counting I/O (similar to RocksDB stats tickers), that could be used for a unit test to check prefix filtering works correctly.

PR #151, opened 2025-09-01
Add prefix bloom filter support

zaidoon1:


src/segment/mod.rs
```diff
@@ -344,18 +417,25 @@ impl Segment {
...
+        // Handle bloom filter skip case
+        if skip_by_bloom {
+            #[cfg(feature = "metrics")]
+            self.metrics
+                .bloom_filter_hits
+                .fetch_add(1, std::sync::atomic::Ordering::Relaxed);
+
+            // Set invalid bounds to ensure no items are yielded
+            iter.set_lower_bound(UserKey::from(&[255u8; 1]));
+            iter.set_upper_bound(UserKey::from(&[0u8; 1]));
```
marvin-j97: Maybe if the function returns an `Option<Iterator>`, we don't need this trickery.
`create_range` (in src/range.rs) and wherever else, can then just flatten out `None`s.

src/segment/mod.rs
```diff
@@ -315,6 +381,14 @@ impl Segment {
         use crate::fallible_clipping_iter::FallibleClippingIter;
         use block_index::iter::create_index_block_reader;
 
+        // Check bloom filter for prefix existence if we have a start bound
+        let skip_by_bloom = if let Bound::Included(key) | Bound::Excluded(key) = range.start_bound()
```
marvin-j97: Also this does not work:

Consider we store [b, c] and use full prefixes.
We query for a..=c.
"a" is extracted and does not exist (bloom true positive). The segment will be skipped and we don't get b, c back.

One possible way could be to maybe check that start and end bound have a common prefix that is being extracted, so:

```rs
let common_prefix = get_common_prefix(start, end);

if !filter.has_prefix(common_prefix, extractor) {
  return None;
}
```

Though it would probably be more efficient to do this just once in `create_range` and then filter by something like !`Segment::contains_prefix(common_prefix, extractor)`.

src/segment/mod.rs
zaidoon1: updated it to be `Option<Iterator>`

src/segment/mod.rs
zaidoon1: alright, I believe I've fixed this up now and added more tests to cover various edge cases. I've kept it as a separate commit so you can see what I changed/new tests added

src/segment/mod.rs
```diff
@@ -296,11 +319,169 @@ impl Segment {
...
+                    if start_prefix == end_prefix {
+                        Some(start_prefix.to_vec())
+                    } else {
+                        // Different prefixes, no common prefix
+                        None
+                    }
+                } else {
+                    None
+                }
+            }
```
marvin-j97: Instead of running one extractor over each bound, getting the common prefix and extracting over that should yield the same result, but need less work (and simpler code).

```rs
let common_prefix = get_common_prefix(start_bound, end_bound);
// lo = abcdef
// hi = abcdxy
// common_prefix = abcd

extract_prefix(common_prefix)
// ...
```

src/segment/mod.rs
```diff
@@ -296,11 +319,169 @@ impl Segment {
...
+            (Some(key), None) | (None, Some(key)) => {
+                // If only one bound exists, use its extracted prefix
+                let prefixes: Vec<_> = prefix_extractor.extract(key).collect();
+                prefixes.first().map(|p| p.to_vec())
```
marvin-j97: There is a lot of collect() and first() here and there... why not just next()? So we avoid the vector allocation.

src/segment/writer/mod.rs
```diff
@@ -153,8 +164,24 @@ impl Writer {
...
+                    let mut prefixes = extractor.extract(&item.key.user_key);
+                    if let Some(first) = prefixes.next() {
+                        // Key is in domain, add first prefix
+                        self.bloom_hash_buffer.push(Builder::get_hash(first));
+                        // Add remaining prefixes
+                        for prefix in prefixes {
+                            self.bloom_hash_buffer.push(Builder::get_hash(prefix));
+                        }
+                    }
+                    // If no prefixes (out of domain), don't add to bloom filter
```
marvin-j97: Why check for the first prefix separately? Instead of just:

```rs
for prefix in extractor.extract(&item.key.user_key) {
    self.bloom_hash_buffer.push(Builder::get_hash(prefix));
}
```

src/segment/mod.rs
zaidoon1: I don't think that would work? For example:

let's say we have a custom prefix extractor that determines the prefix length based on the suffix of the key:

```
struct SuffixAwareExtractor;
impl PrefixExtractor for SuffixAwareExtractor {
    fn extract<'a>(&self, key: &'a [u8]) -> Box<dyn Iterator<Item = &'a [u8]> + 'a> {
        // If key ends with a digit, extract first 2 chars
        // If key ends with a letter, extract first 3 chars
        if key.len() >= 3 {
            if key[key.len() - 1].is_ascii_digit() {
                Box::new(std::iter::once(&key[..2]))
            } else {
                Box::new(std::iter::once(&key[..3]))
            }
        } else if key.len() >= 2 {
            Box::new(std::iter::once(&key[..2]))
        } else {
            Box::new(std::iter::empty())
        }
    }
}
```

Now let's say we are doing:

Range: "abc1".."abcz"

current code:
Extract from "abc1" → "ab" (ends with digit '1', so takes first 2 chars)
Extract from "abcz" → "abc" (ends with letter 'z', so takes first 3 chars)
Compare: "ab" != "abc"
Returns: None


Suggested code (if I'm understanding correctly):
Common bytes of "abc1" and "abcz" → "abc" (3 chars)
Extract from "abc" → "abc" (ends with letter 'c', so takes first 3 chars)
Returns: Some("abc")


In this case, we produce a different output. My current code returns None correctly identifying that the range spans different prefixes, while the suggested code incorrectly returns Some("abc"). This would cause the bloom filter optimization to incorrectly skip segments when searching for the range "abc1".."abcz", potentially missing valid results that start with "ab" but not "abc".

The fundamental issue is that prefix extractors may need the full key context to determine the correct prefix - they're not just simple byte slicers. The extractor's logic might depend on patterns, delimiters, or other characteristics that appear anywhere in the key.

Please let me know if I'm misunderstanding something

src/run_reader.rs
marvin-j97, line 1: There will probably need to be some unit tests for the run reader, especially for L1.

Ideally in [`create_range`](https://github.com/zaidoon1/lsm-tree/blob/5e53adb36a4c59b424b67498d87e7fc67a157414/src/range.rs#L242-L254) we can completely rule out the run if all eligible segments don't contain a prefix extractor's prefixes.
However, for larger scans, that may get really expensive to check every segment, so maybe it needs to be performed lazily when the next segment iterator in a run is setup.

If we just skip the first segment and then read the next, that still incurs 1 I/O, so the prefix extraction didn't really help in that case.

src/segment/filter/blocked_bloom/mod.rs
```diff
@@ -128,6 +128,30 @@ impl<'a> BlockedBloomFilterReader<'a> {
     pub fn contains(&self, key: &[u8]) -> bool {
         self.contains_hash(Self::get_hash(key))
     }
+
+    /// Returns `true` if any prefix of the key may be contained.
+    ///
+    /// Returns `None` if the key is out of domain.
+    #[must_use]
+    #[inline]
```
marvin-j97: I generally don't put inline statements because LLVM is pretty aggressive about it anyway except for when I can actually prove that it is not inlined normally (e.g. in godbolt), so I would just remove those.

src/metrics.rs
```diff
@@ -45,4 +45,14 @@ impl Metrics {
...
         hits / queries
     }
+
+    /// Number of bloom filter queries performed.
+    pub fn bloom_filter_queries(&self) -> usize {
+        self.bloom_filter_queries.load(Relaxed)
+    }
+
+    /// Number of bloom filter hits (queries that avoided disk I/O).
+    pub fn bloom_filter_hits(&self) -> usize {
```
marvin-j97: `bloom_filter_negatives` or `io_skipped_by_filter` maybe; `hits` sounds like a `true positive`, but really the bloom filter is trying to find a `true negative`.

src/metrics.rs
```diff
@@ -45,4 +45,14 @@ impl Metrics {
         let hits = self.bloom_filter_hits.load(Relaxed) as f64;
         hits / queries
     }
+
+    /// Number of bloom filter queries performed.
+    pub fn bloom_filter_queries(&self) -> usize {
```
marvin-j97: Also, I named this wrong in `bloom_filter_efficiency`, but it should be called `filter_*`, there may be other types than Bloom filters.

src/prefix.rs
```diff
@@ -0,0 +1,300 @@
...
+///     fn name(&self) -> &str {
+///         "hierarchical_prefix"
+///     }
+/// }
+///
+/// // Example 2: Domain-based extractor for email-like keys
+/// // For "user@example.com", generates:
+/// // - "example.com" (domain)
+/// // - "user@example.com" (full key)
+/// struct EmailDomainExtractor;
```
marvin-j97: I don't think this is a realistic example because we can never perform a prefix scan over the domain.
More realistic would be an flipped key, e.g. "example.com@user" - that way you can scan over `example.com@`. That would extract `example.com`, while a point read to "example.com@user" can use the normal bloom filter with the full key.

Remember that the key is still sorted lexicographically, so suffix extraction does not really make sense.

src/segment/writer/mod.rs
zaidoon1: fixed

src/run_reader.rs
zaidoon1: done, let me know if I misunderstood

src/segment/filter/blocked_bloom/mod.rs
zaidoon1: removed

src/prefix.rs
zaidoon1: changed example

src/segment/mod.rs
zaidoon1: fixed

src/segment/util.rs
```diff
@@ -13,48 +13,58 @@ use std::{path::Path, sync::Arc};
 #[derive(Debug)]
 pub struct SliceIndexes(pub usize, pub usize);
 
-pub fn load_block(
```
marvin-j97, line 19: Changing this will probably infer with what I'm doing a bit (and it seemingly already did), adding this in a new PR would be better.

zaidoon1: reverted changes for now

src/segment/filter/blocked_bloom/mod.rs
zaidoon1: need to fix this, this was re-added with the revert

src/run_reader.rs
zaidoon1: this was reverted to avoid the merge conflicts. Do you want me to do this in this PR or could this be a follow up?

src/segment/filter/blocked_bloom/mod.rs
zaidoon1: fixed for real now

src/prefix.rs
zaidoon1: let me fix this one too again

src/segment/mod.rs
```diff
@@ -296,11 +319,154 @@ impl Segment {
...
     #[doc(hidden)]
-    pub fn iter(&self) -> impl DoubleEndedIterator<Item = crate::Result<InternalValue>> {
+    pub fn iter(&self) -> Option<impl DoubleEndedIterator<Item = crate::Result<InternalValue>>> {
         self.range(..)
     }
 
+    /// Returns true if the prefix bloom filter indicates the prefix doesn't exist.
+    /// This is used to potentially skip segments during range queries.
+    /// Only works when a prefix extractor is configured.
+    fn should_skip_by_prefix_bloom_filter(&self, key: &UserKey) -> bool {
```
marvin-j97: `key: &[u8]` to not require an owned type should work here

zaidoon1: fixed

src/prefix.rs
zaidoon1: fixed

src/run_reader.rs
zaidoon1: pushed a separate commit for this, let me know what you think

src/run_reader.rs
```diff
@@ -29,6 +29,24 @@ impl RunReader {
...
+        let mut has_potential_match = false;
+
+        for segment in segments_in_range {
+            // Check if segment might contain data for this range
+            // If segment has no prefix extractor or can't determine, assume it might contain data
+            if segment.might_contain_range(&range) {
+                has_potential_match = true;
+                break;
+            }
+        }
```
marvin-j97: Need to benchmark this in a run with 100 - 1000 segments.
If we have a large scan over the entire keyspace, and even if `might_contain_range` runs in just ~10ns, that can easily be 1-10µs of CPU work.
In reality I don't know how likely a large level may be completely filtered out by prefix filters, but it's something to keep in mind.
We may also be able to skip this loop completely if we don't have a prefix extractor configured at all (which most applications probably won't), but we would need access to it somehow by function parameter.

zaidoon1: i added some benchmarking code and capped the number of scans, also skipped this for when prefix bloom is not set. Here is my run locally on m4 mac pro:

```
=== RunReader Performance Benchmark ===
Testing impact of prefix bloom filter checks on large runs


--- Testing with 100 segments ---
  Without prefix extractor:      420 ns/query
  With prefix extractor:         451 ns/query
  Overhead: 31 ns (7.4%)
  Per-segment overhead: ~0 ns

--- Testing with 500 segments ---
  Without prefix extractor:      441 ns/query
  With prefix extractor:         445 ns/query
  Overhead: 4 ns (0.9%)
  Per-segment overhead: ~0 ns

--- Testing with 1000 segments ---
  Without prefix extractor:      410 ns/query
  With prefix extractor:         418 ns/query
  Overhead: 8 ns (2.0%)
  Per-segment overhead: ~0 ns
  ```

File: src/metrics.rs (hunk 8 above)
Comment (zaidoon1) [reply src/metrics.rs]: removed as many references of "bloom"  as possible to make it generic

File: src/metrics.rs (hunk 7 above)
Comment (zaidoon1) [reply src/metrics.rs]: updated the naming to be less confusing

File: src/config.rs [hunk 13]
```diff
@@ -312,6 +320,30 @@ impl Config {
         self
     }
 
+    /// Sets the prefix extractor for bloom filters.
```
marvin-j97:
```suggestion
    /// Sets the prefix extractor for filters.
```
etc.

src/run_reader.rs
```diff
@@ -29,6 +29,45 @@ impl RunReader {
...
+        // (all segments in a run should have the same configuration)
+        let has_prefix_extractor = run
+            .get(lo)
+            .map(|seg| seg.has_prefix_extractor())
+            .unwrap_or(false);
+
+        if has_prefix_extractor {
+            // Only perform prefix filter checks if a prefix extractor is configured
+            // This avoids unnecessary CPU work for large scans when prefix filtering isn't used
+            let segments_in_range = run.get(lo..=hi).ok_or(()).ok()?;
```
marvin-j97:
```suggestion
            let segments_in_range = run.get(lo..=hi)?;
```

src/config.rs
zaidoon1: done

src/run_reader.rs
zaidoon1: fixed

zaidoon1:
did i break the build and test lsm example build or some other isssue?

marvin-j97:
> did i break the build and test lsm example build or some other isssue?

The CI examples script does not support raw .rs files in the examples/ folder, so it crashes.

zaidoon1:
> > did i break the build and test lsm example build or some other isssue?
> 
> The CI examples script does not support raw `.rs` files in the `examples/` folder, so it crashes.

should be fixed now, thanks!

src/prefix.rs
```diff
@@ -0,0 +1,300 @@
...
+pub trait PrefixExtractor: Send + Sync {
+    /// Extracts zero or more prefixes from a key.
+    ///
+    /// All prefixes will be added to the filter during segment construction.
+    ///
+    /// An empty iterator means the key is "out of domain" and won't be added to the filter.
+    fn extract<'a>(&self, key: &'a [u8]) -> Box<dyn Iterator<Item = &'a [u8]> + 'a>;
+
+    /// Returns a unique name for this prefix extractor.
+    fn name(&self) -> &str;
```
marvin-j97, line 68: I'm thinking how to make use of this.
I guess this should be stored in each keyspace's config (in fjall) to check that the given prefix extractor matches the one that was initially configured.

zaidoon1: right, I may need to do more work but in rocksdb, if you want to change your prefix extractor then you need to use a different name which would cause rocksdb to ignore the old bloom that was used for the existing files

marvin-j97: Then maybe the prefix extractor name also needs to be stored in every table's properties.

zaidoon1: I've pushed a commit  to do this and match rocksdb's behaviour:

"If prefix extractor changes when DB restarts, some SST files may contain prefix bloom filters generated using different prefix extractor than the one in current options. When opening those files, RocksDB will compare prefix extractor name stored in the properties of the SST files. If the name is different from the prefix extractor provided in the options, the prefix bloom filter is not used for this file"


one thing is how does fjall handle options that can be dynamically changeable vs can't be changed without a db restart? I don't think we would want a user to be able to change a prefix extractor on the fly

marvin-j97: Currently the configuration cannot be changed during runtime. However I don't think it would be a problem, because tables that use the old prefix extractor would continue using that, and otherwise the filter is skipped. But that's probably an edge case and just shouldn't be allowed.

zaidoon1: ok cool, so then what I have here should be good.

PR #155, opened 2025-09-17
Add zlib compression feature

harshithsaiv:
This PR adds support for `Zlib(level)` gated by the optional `zlib` Cargo feature. 

## Changes

- **CompressionType**
  - New variant: `Zlib(u8)`.
  - Updated `Encode`/`Decode`:
    - Levels are clamped to `0..=9` on encode/decode to match `flate2` expectations.

- **Block I/O**
  - Zlib branches added in `write_into`, `from_reader`, and `from_file`.
  - Compression via `flate2::write::ZlibEncoder`.
  - Decompression via `flate2::read::ZlibDecoder`.

- **Blob path**
  - `blob_tree/compression.rs`: `MyCompressor` added support for zlib
  - `blob_tree/mod.rs`: `BlobTree::open()` wires `CompressionType::{None, Lz4, Zlib(_)}` .

- **Cargo.toml**
  - Add optional dependency: `flate2` with `zlib-rs` backend (`default-features = false`).
  - Introduce `zlib` feature that enables the dependency.

- No breaking changes when i ran cargo build --features zlib  and cargo test --features zlib 

One open point: I clamped levels ≥10 to 9 during encode. Would you prefer rejecting invalid levels instead?

### DIFF
--- a/src/segment/block/mod.rs
+++ b/src/segment/block/mod.rs
@@ -224,7 +224,7 @@ impl Block {
                     unsafe { Slice::builder_unzeroed(header.uncompressed_length as usize) };
                 d.read_exact(&mut decompressed_data)
                     .map_err(|_| crate::Error::Decompress(compression))?;
-                decompressed_data.freeze().into
+                decompressed_data.freeze().into()
             }
         };
 


src/blob_tree/compression.rs
marvin-j97, line 1: This file will be removed (soon), so the changes here can be reverted

src/blob_tree/mod.rs
marvin-j97, line 1: Same here, the Compressor trait will not exist anymore, so this can be reverted

src/segment/block/mod.rs
```diff
@@ -56,16 +60,24 @@ impl Block {
             previous_block_offset: BlockOffset(0), // <-- TODO:
         };
 
-        let data = match compression {
-            CompressionType::None => data,
+        let data: Vec<u8> = match compression {
+            CompressionType::None => data.to_vec(),
```
marvin-j97: `.to_vec()` causes an additional heap allocation

src/segment/block/mod.rs
```diff
@@ -100,6 +112,15 @@ impl Block {
...
                 builder.freeze().into()
             }
+
+            #[cfg(feature = "zlib")]
+            CompressionType::Zlib(_level) => {
+                let mut d = ZlibDecoder::new(&raw_data[..]);
+                let mut decompressed_data = Vec::with_capacity(header.uncompressed_length as usize);
+                d.read_to_end(&mut decompressed_data)
+                    .map_err(|_| crate::Error::Decompress(compression))?;
+                Slice::from(decompressed_data)
```
marvin-j97: Vec will be heap allocated, then the Slice::from can cause an additional heap allocation + memcpy + buffer zeroing.
It would be better to use a SliceBuilder as seen in the Lz4 match arm.

src/segment/block/mod.rs
```diff
@@ -191,6 +212,17 @@ impl Block {
...
+
+            #[cfg(feature = "zlib")]
+            CompressionType::Zlib(_level) => {
+                #[allow(clippy::indexing_slicing)]
+                let raw_data = &buf[Header::serialized_len()..];
+                let mut d = ZlibDecoder::new(raw_data);
+                let mut decompressed_data = Vec::with_capacity(header.uncompressed_length as usize);
+                d.read_to_end(&mut decompressed_data)
+                    .map_err(|_| crate::Error::Decompress(compression))?;
+                byteview::ByteView::from(decompressed_data)
```
marvin-j97: Also use SliceBuilder

src/compression.rs
```diff
@@ -34,6 +39,13 @@ impl Encode for CompressionType {
             Self::Lz4 => {
                 writer.write_u8(1)?;
             }
+
+            #[cfg(feature = "zlib")]
+            Self::Zlib(level) => {
+                writer.write_u8(2)?;
+                let lvl = (*level).min(9);
```
marvin-j97: Rejecting would probably be better than just clamping.

Not sure what error enum variant to use here, probably needs a new one (DecodeError::InvalidCompressionLevel(...)).

marvin-j97 (changes requested):

src/blob_tree/compression.rs
harshithsaiv: Reverted the changes !

src/blob_tree/mod.rs
harshithsaiv: Reverted the changes !

src/segment/block/mod.rs
harshithsaiv: Should i make use of borrow concept borrow::Cow ? so that there is no allocations or write per-branch and early-return which one should I implement ?

src/segment/block/mod.rs
harshithsaiv: I have changed it and have used SliceBuffer now

src/segment/block/mod.rs
harshithsaiv: made the change

src/compression.rs
harshithsaiv: Okay I will change it to rejection and handle that error rather than clamping .

src/segment/block/mod.rs
```diff
@@ -56,16 +61,24 @@ impl Block {
...
             #[cfg(feature = "lz4")]
-            CompressionType::Lz4 => &lz4_flex::compress(data),
+            CompressionType::Lz4 => Cow::Owned((lz4_flex::compress(data))),
+
+            #[cfg(feature = "zlib")]
+            CompressionType::Zlib(level) => {
+                let lvl = level as u32;
+                let mut e = ZlibEncoder::new(Vec::new(), ZCompression::new(lvl));
+                e.write_all(data)?;
+                Cow::Owned(e.finish()?)
```
marvin-j97, line 75: I think it should work by just returning as a reference `&e.finish()?`, as I did before - but using Cow shouldn't have any meaningful negative impact I think

src/coding.rs
```diff
@@ -57,6 +58,8 @@ pub enum DecodeError {
 
     /// UTF-8 error
     Utf8(std::str::Utf8Error),
+
+    InvalidCompressionLevel(u8),
```
marvin-j97: If we ever have zstd support in the future we would need a different enum type because zstd levels are... weird (I think they need an i32).

So the obvious choice would be to just use an i32 always, and maybe make it a tuple: `InvalidCompressionLevel((CompressionType, i32))`.

harshithsaiv:
Hi @marvin-j97 Marvin,
I’m having trouble viewing your latest comment in the PR GitHub isn’t showing it to me. Could you please repost or clarify your feedback so I can address it? Thank you!

marvin-j97 (changes requested):

marvin-j97:
Oops the review was not sent

src/segment/block/mod.rs
harshithsaiv: I can return `e.finish()?` directly and have the return type as Vec<u8> but that would require `.to_vec()` for the uncompressed case and you suggested that it would create unnecessary heap allocation so i went on implementing the `Cow` approach

src/coding.rs
harshithsaiv: I changed  the InvalidCompressionLevel() `EncodeError`

marvin-j97 (commented):
Just rustfmt missing I think

Issue #148, opened 2025-08-21
DropRange


marvin-j97:
Drop all segments that are fully contained in a key range.
Is not transactional.

marvin-j97:
The drop range function exists now, but the API needs to be replaced with a impl RangeBounds (if possible).

PR #156, opened 2025-09-27
update drop_range API to use RangeBounds

zaidoon1:
to address https://github.com/fjall-rs/lsm-tree/issues/148#issuecomment-3341845048

### DIFF
--- a/src/compaction/drop_range.rs
+++ b/src/compaction/drop_range.rs
@@ -33,38 +33,6 @@ impl RangeBounds<Slice> for OwnedBounds {
     }
 }
 
-struct ContainedSegments<'a> {
-    segments: &'a [Segment],
-    bounds: &'a OwnedBounds,
-    pos: usize,
-}
-
-impl<'a> ContainedSegments<'a> {
-    fn new(segments: &'a [Segment], bounds: &'a OwnedBounds) -> Self {
-        Self {
-            segments,
-            bounds,
-            pos: 0,
-        }
-    }
-}
-
-impl<'a> Iterator for ContainedSegments<'a> {
-    type Item = &'a Segment;
-
-    fn next(&mut self) -> Option<Self::Item> {
-        while let Some(segment) = self.segments.get(self.pos) {
-            self.pos += 1;
-
-            if self.bounds.contains(segment.key_range()) {
-                return Some(segment);
-            }
-        }
-
-        None
-    }
-}
-
 impl OwnedBounds {
     #[must_use]
     pub fn contains(&self, range: &KeyRange) -> bool {
@@ -93,10 +61,6 @@ pub struct Strategy {
 
 impl Strategy {
     /// Configures a new `DropRange` compaction strategy.
-    ///
-    /// # Panics
-    ///
-    /// Panics, if `target_size` is below 1024 bytes.
     #[must_use]
     #[allow(dead_code)]
     pub fn new(bounds: OwnedBounds) -> Self {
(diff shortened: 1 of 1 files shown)


src/abstract.rs
```diff
@@ -78,10 +78,13 @@ pub trait AbstractTree {
 
     /// Drops segments that are fully contained in a given range.
     ///
+    /// Both range bounds must be inclusive and finite.
```
marvin-j97: If we have tables like this:

```
[a b c] [d e f] [g h i]
1       2       3
```

Dropping `..g` should drop 1 & 2, etc. So I don't see why certain range bounds should not be supported (unless I'm missing something).

And `..` should drop the entire tree (basically a clear()).

zaidoon1: We can definitely support that, i wasn't sure what you had in mind in terms of whether to support just an upper bound or no bounds at all to (clear). I can make the change to let the user do whatever they want.

zaidoon1: I'll assume the only thing that we want to reject is when lower range is higher than upper range

zaidoon1: done, let me know what you think

marvin-j97: I think it would be enough to early return if lo > hi (treat it as an empty range).

zaidoon1: done

zaidoon1:
linter failure is related to:
```
error: the variable `count` is used as a loop counter
   --> src/tree/mod.rs:149:9
    |
149 |         for (key, value) in iter {
    |         ^^^^^^^^^^^^^^^^^^^^^^^^ help: consider using: `for (count, (key, value)) in iter.enumerate()`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#explicit_counter_loop
    ```
    but I did't touch this code?

**Comment by marvin-j97** [comment]
> linter failure is related to:
> 
> ```
> error: the variable `count` is used as a loop counter
>    --> src/tree/mod.rs:149:9
>     |
> 149 |         for (key, value) in iter {
>     |         ^^^^^^^^^^^^^^^^^^^^^^^^ help: consider using: `for (count, (key, value)) in iter.enumerate()`
>     |
>     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#explicit_counter_loop
> 
>     
> ```
> 
> ```
> but I did't touch this code?
> ```

I think the linter rules changed again - I updated the 3.0.0 branch to allow it.

File: src/tree/mod.rs [hunk 2]
```diff
@@ -163,14 +164,17 @@ impl AbstractTree for Tree {
 
         visible_seqno.fetch_max(seqno + 1);
 
-        log::info!("Ingested {count} items in {:?}", start.elapsed());
```
zaidoon1, line 167: accidentally removed this, going to restore it

src/compaction/drop_range.rs
```diff
@@ -34,7 +64,10 @@ impl CompactionStrategy for Strategy {
             .current_version()
             .iter_levels()
             .flat_map(|lvl| lvl.iter())
-            .flat_map(|run| run.get_contained(&self.key_range))
+            .flat_map(|run| {
+                run.iter()
+                    .filter(|segment| self.bounds.contains(segment.key_range()))
+            })
```
marvin-j97, line 87: It's a bit unfortunate to lose `Run::get_contained` here because that actually uses binary search instead of a linear filter, which is nicer when dropping small ranges from large levels (think 1000+ tables).
So that might be worth bringing back, as the compaction strategies run while holding a write lock, so they need to be somewhat quick.

zaidoon1: updated the code to use the binary search code path for initial filtering

marvin-j97 (approved):

Issue #171, opened 2025-10-20
Try to make iterators 'static again


marvin-j97:
Currently the iterators are lifetime-bound to the trees, because in the BlobTree, each Guard needs to get a Version reference (tree.current_version).

Can we clone the version handle so we own it and store it in the iterator return value, and (if needed, self cell could be used), give a reference of that to the Guards.

That would also avoid the Arc Clone for every KV we visited, instead it's just one clone once of the Version.

Because the iterator holds the Version, we cannot lose data while it is open.

This has the consequence of relaxing the iterator API of transactions in fjall as it was before 3.0.0, which is more ergonomic.

PR #173, opened 2025-10-22
make iterators Send + 'static to support cross-thread usage

zaidoon1:
to resolve https://github.com/fjall-rs/lsm-tree/issues/171


src/blob_tree/mod.rs
```diff
@@ -180,20 +232,27 @@ impl AbstractTree for BlobTree {
...
                 .create_internal_range(&range, seqno, index)
                 .map(move |kv| {
                     IterGuardImpl::Blob(Guard {
-                        blob_tree: self,
-                        version: version.clone(), // TODO: PERF: ugly Arc clone
+                        tree_id,
+                        blobs_folder: Arc::clone(&blobs_folder),
+                        cache: cache.clone(),
+                        descriptor_table: descriptor_table.clone(),
+                        version: version.clone(),
```
marvin-j97: Note that now we're doing 4 Arc clones per KV.

My idea was to only pass references to each guard because each guard has a lifetime strictly shorter than the iterator itself. I'm not sure if this can be achieved without using `self_cell` or at all though, I have tried a little bit, but I did not decompose the blob tree into its constituent parts as you did.

zaidoon1: so refs won’t work here with the current api because:

1. AbstractTree::{prefix,range} return Box<dyn DoubleEndedIterator<Item = IterGuardImpl> + Send + 'static>.
2. IterGuardImpl and blob_tree::Guard have no lifetime parameters.
3. Standard Iterator cannot yield items that borrow from the iterator state (that would require a lending iterator/GAT-based trait).
4. Therefore, guards must own their dependencies, not borrow. Passing references (even with self_cell) won’t compile under this API . 

Best I can do is to get the arc clones down to 1 per kv

marvin-j97: > IterGuardImpl and blob_tree::Guard have no lifetime parameters

It did work when passing in a &BlobTree (though that bound it to BlobTree's lifetime), so if making it have a lifetime again and that allows to have the state kept in the iterator and lent to the guard, that shouldn't be a problem.

src/blob_tree/mod.rs
```diff
@@ -87,6 +102,47 @@ fn resolve_value_handle(tree: &BlobTree, version: &Version, item: InternalValue)
     }
 }
 
+fn resolve_value_handle_owned(
```
marvin-j97: At this point it would probably be best to remove resolve_value_handle() and only use this function everywhere, otherwise the function is essentially duplicated without much benefit.

And then the _owned suffix can be removed.

zaidoon1: can do

src/blob_tree/mod.rs
```diff
@@ -205,17 +232,21 @@ impl AbstractTree for BlobTree {
...
+        });
 
-        // TODO: PERF: ugly Arc clone
         Box::new(
             self.index
                 .create_internal_range(&range, seqno, index)
                 .map(move |kv| {
                     IterGuardImpl::Blob(Guard {
-                        blob_tree: self,
-                        version: version.clone(), // TODO: PERF: ugly Arc clone
```
marvin-j97, line 217: I just noticed that we can just clone the BlobTree into the Guard.
That would just be 2 Arc clone (fine...), because we still need to hold the Version, but it would avoid the need to heap-allocate the shared state on every iterator setup altogether.

(Actually currently it's 3 Arc clones because the BlobTree does not have an Arc\<Inner\>, but still)

zaidoon1: wouldn't the one small heap allocation per iterator be better than 3 arc clones per kv?

zaidoon1: hmm i guess this all boils down to how many kvs per iterator. frequent iterators, small number of kvs, then heap allocation is bad, but if we have thousands of kvs then it's a different story...

zaidoon1: I can go with what you want and we can see how it goes

Issue #176, opened 2025-10-26
Remove clipping iterator from table range iterators


marvin-j97:
The clipping iterator was added back before the indexes were finished properly. Unsure if the clipping iterator can be removed or if there may be edge cases when seeking.
Either way the clipping iterator adds a lot of key comparisons *per* run that is iterated, so it'a undesirable because it further amplifies the already-high key comparison costs of the LSM-tree.

PR #178, opened 2025-10-29
remove clipping iterator from table range iterators & consolidate key range iteration and bounds handling

zaidoon1:
resolve https://github.com/fjall-rs/lsm-tree/issues/176

### DIFF
--- a/tests/tree_disjoint_point_read.rs
+++ b/tests/tree_disjoint_point_read.rs
@@ -1,5 +1,4 @@
 use lsm_tree::{config::BlockSizePolicy, AbstractTree, Config, SeqNo};
-use std::sync::Arc;
 use test_log::test;
 
 #[test]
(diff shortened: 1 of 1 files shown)


src/table/iter.rs
```diff
@@ -20,6 +20,9 @@ use crate::metrics::Metrics;
 
 type InnerIter<'a> = DataBlockIter<'a>;
 
+type Bound = (UserKey, bool);
```
marvin-j97: An enum similar to std::ops::Bound but without Unbounded would probably be the more readable solution here

zaidoon1: fixed

src/table/iter.rs
```diff
@@ -312,15 +352,20 @@ impl DoubleEndedIterator for Iter {
...
             }
-            if let Some(key) = &self.range.0 {
-                // Apply the lower bound as well so that we never step past the low key when
-                // iterating backwards through the block.
-                reader.seek_lower(key, SeqNo::MAX);
+            if let Some(bound) = &self.range.0 {
+                match bound {
+                    Bound::Excluded(key) => reader.seek_lower_exclusive(key, SeqNo::MAX),
+                    Bound::Included(key) => reader.seek_lower(key, SeqNo::MAX),
+                };
```
marvin-j97: I think it's great to support both bound types in the block reader directly.
Though if we made `reader.seek_lower_exclusive` private and rename `reader.seek_lower` to `reader.seek_lower_inclusive`, we could group this to be `seek(Bound)`, so that the caller does not have to match on its bound key. That would simplify the code in iter and table.

zaidoon1: pushed a commit to do this

marvin-j97 (approved):

PR #179, opened 2025-10-30
implement TTL support for FIFO compaction

zaidoon1:
resolve https://github.com/fjall-rs/lsm-tree/issues/177

### DIFF
--- a/src/table/inner.rs
+++ b/src/table/inner.rs
@@ -15,7 +15,7 @@ use crate::{
 };
 use std::{
     path::PathBuf,
-    sync::{atomic::AtomicBool, Arc},
+    sync::{atomic::AtomicBool, Arc, OnceLock},
 };
 
 pub struct Inner {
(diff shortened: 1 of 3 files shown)


src/table/inner.rs
```diff
@@ -55,6 +55,10 @@ pub struct Inner {
 
     #[cfg(feature = "metrics")]
     pub(crate) metrics: Arc<Metrics>,
+
+    /// Cached sum of referenced blob file bytes for this table.
+    /// Lazily computed on first access to avoid repeated I/O in compaction decisions.
+    pub(crate) cached_blob_bytes: OnceLock<u64>,
```
marvin-j97, line 61: Nice! I was thinking of just loading the blob refs on table recovery instead of doing I/O, but I think OnceLock is the best of both worlds

marvin-j97:
Thanks, I think that mostly concludes the remaining work for lsm-tree v3.

PR #186, opened 2025-11-03
Add prefix filter support

zaidoon1:
rebased https://github.com/fjall-rs/lsm-tree/pull/151 on top of main and adapted various things to the new api

## Summary by CodeRabbit

* **New Features**
  * Prefix extractors for prefix-aware filtering, persisted extractor metadata, and compatibility checks; reads can skip non-matching table ranges to reduce work; writers record/use extractors so filters index prefixes

* **Tests**
  * Extensive unit, integration and recovery tests covering extractor behavior, compatibility, and range-skipping

* **Fuzz**
  * New fuzz harness validating prefix-filter correctness; improved fuzz run setup and ignore rules

* **Benchmarks**
  * Benchmarks measuring range-query performance with and without extractors

* **Chores**
  * Robust directory creation in fuzz run instructions (mkdir -p)


marvin-j97:
Unfortunately this has been hit by another wave of conflicts, but I just released 3.0.0, so there will be a bit of a freeze of activity from this point on.

src/config/mod.rs
```diff
@@ -308,6 +315,14 @@ impl Config {
         }
     }
 
+    /// Sets the prefix extractor for building prefix-aware Bloom filters.
```
marvin-j97:
```suggestion
    /// Sets the prefix extractor for building prefix-aware filters.
```

tests/prefix_filter.rs
marvin-j97, line 1: This file is probably at the point where it could be split into multiple smaller files. But I can do that later.

src/table/writer/mod.rs
```diff
@@ -92,6 +92,10 @@ pub struct Writer {
     linked_blob_files: Vec<LinkedFile>,
 
     initial_level: u8,
+
+    /// Optional prefix extractor used to register extracted prefixes into the Bloom filter.
```
marvin-j97:
```suggestion
    /// Optional prefix extractor used to register extracted prefixes into the filter.
```

src/config/mod.rs
```diff
@@ -226,6 +226,11 @@ pub struct Config {
     /// Filter construction policy
     pub filter_policy: FilterPolicy,
 
+    /// Optional prefix extractor used to construct prefix-aware filters.
+    /// When set, the table writer will add extracted prefixes (instead of full keys)
+    /// to the Bloom filter and persist the extractor name in table metadata for compatibility checks.
```
marvin-j97:
```suggestion
    /// to filters and persist the extractor name in table metadata for compatibility checks.
```

src/table/mod.rs
```diff
@@ -605,6 +699,67 @@ impl Table {
...
+            std::ops::Bound::Unbounded => None,
+        };
+
+        let start_pref = start_key.and_then(|k| extractor.extract(k).next());
+        let end_pref = end_key.and_then(|k| extractor.extract(k).next());
+
+        if let (Some(sp), Some(ep)) = (start_pref, end_pref) {
+            if sp == ep {
+                if let Some(sk) = start_key {
+                    if matches!(self.maybe_contains_prefix(sk, extractor), Ok(Some(false))) {
```
marvin-j97: `maybe_contains_prefix` again checks the prefix extractor name even though we already validated it above.

src/table/mod.rs
```diff
@@ -605,6 +699,67 @@ impl Table {
...
+                }
+                return false;
+            }
+        }
+
+        if let Some(sp) = start_pref {
+            if let Some(mp) = extractor.extract(self.min_key().as_ref()).next() {
+                if sp == mp {
+                    if let Some(sk) = start_key {
+                        if matches!(self.maybe_contains_prefix(sk, extractor), Ok(Some(false))) {
```
marvin-j97: ... and possibly again.

src/run_reader.rs
```diff
@@ -15,46 +18,151 @@ pub struct RunReader {
...
+        Some(Self::culled(run, range, (Some(lo), Some(hi)), extractor))
     }
 
+    /// Creates a run reader with precomputed overlap indices.
+    ///
+    /// This variant assumes the caller already determined the overlapping table
+    /// indices. It initializes boundary table readers and
+    /// performs lazy per-table prefix-filter skipping during iteration.
     #[must_use]
     pub fn culled<R: RangeBounds<UserKey> + Clone + Send + 'static>(
```
marvin-j97: I would recommend running the mutation tests and see if any of these functions pop up in it. Because if there's a single wrong edge case in there it's easy to mess up the read path.
In fact, prefix extractors are so invasive to the read path that it may be worth thinking about fuzz testing them (so we don't get wrongly applied filters or anything).

zaidoon1: ran the mutations tests, they didnt' find real bugs, but I added more tests to cover the "missed" cases

zaidoon1: i've added a test file that does fuzz testing so that it runs on each PR. It found a bug and I fixed it as well.

marvin-j97: Could the fuzz be moved into the fuzz folder with cargo-afl maybe? Then it can be run in an infinite loop, too, and AFL will save all crashes to rerun deterministically.

zaidoon1: done

zaidoon1:
i've just pushed a commit, this adds coverage for the last valid case that i missed. The rest is "safe fallbacks" for code that should not be hit. What would prefer? I've added debug_asserts and kept the safe fallback. I can also expect/panic all together if you prefer?

marvin-j97:
Found a couple crashes through fuzz testing at 77d0911bc97b143138c5825fd65e05bb18de1b66:

[crashes_1.tar.gz](https://github.com/user-attachments/files/25376014/crashes_1.tar.gz)

zaidoon1:
> > i've just pushed a commit, this adds coverage for the last valid case that i missed. The rest is "safe fallbacks" for code that should not be hit. What would prefer? I've added debug_asserts and kept the safe fallback. I can also expect/panic all together if you prefer?
> 
> It's probably fine. The fallback is a bit unfortunate because it just requires an additional range checking - I guess one could just trust the `range_overlap_indexes` function, on the other hand, maybe just removing the safe guard may be fine as well? If `range_overlap_indexes` were to actually return a non-overlapping table, would this be critical?

I've removed the fallback, kept the debug_assert and added comments on what would happen if a refactor broke the code and it wasn't caught by the debug_assert (i.e only wasted I/O)

src/table/writer/filter/partitioned.rs
```diff
@@ -176,6 +180,24 @@ impl<W: std::io::Write + std::io::Seek> FilterWriter<W> for PartitionedFilterWri
...
+        self.approx_filter_size = self
+            .bloom_policy
+            .estimated_filter_size(self.bloom_hash_buffer.len());
+
+        if self.approx_filter_size >= self.partition_size as usize {
+            // last_key is set by either register_key or notify_key.
+            // It should always be set before register_bytes is called.
+            if let Some(key) = self.last_key.clone() {
+                self.spill_filter_partition(&key)?;
+            }
```
marvin-j97: This change resulted in a regression of coverage - the spilling in register_key is covered by tests, but this code path isn't. So somewhere is another edge case in filter writing that is not considered in unit tests.

zaidoon1: fixed

tests/prefix_filter_recovery.rs
```diff
@@ -0,0 +1,196 @@
...
+
+        assert_eq!(keys.len(), 1);
+
+        #[cfg(feature = "metrics")]
+        {
+            let final_queries = tree.metrics().filter_queries();
+
+            // After recovery, filters should still be working
+            assert!(final_queries > 0, "filter should not be used");
+        }
```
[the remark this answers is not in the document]
zaidoon1: fixed

src/table/mod.rs
```diff
@@ -226,6 +298,23 @@ impl Table {
         self.metadata.file_size
     }
 
+    pub(crate) fn get_without_filter(
+        &self,
+        key: &[u8],
+        seqno: SeqNo,
+    ) -> crate::Result<Option<InternalValue>> {
+        self.point_read(key, seqno)
+    }
```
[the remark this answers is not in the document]
zaidoon1: this bug is currently unreachable through the public API because:
1. global_seqno != 0 only occurs for ingested tables
2. ingested entries always have internal seqno 0
3. Version-level MVCC filtering (get_version_for_snapshot) ensures the ingested table is only visible when version.seqno < query_seqno, which guarantees global_seqno < query_seqno
4. Under that invariant, both the normalized and unnormalized paths produce identical results

however i've fixed it in case things change in the future

fuzz/prefix_filter/src/main.rs
```diff
@@ -0,0 +1,644 @@
...
+// Helpers
+// ---------------------------------------------------------------------------
+
+fn collect_kv(iter: impl Iterator<Item = IterGuardImpl>) -> Vec<(Vec<u8>, Vec<u8>)> {
+    iter.filter_map(|g| {
+        let (k, v) = g.into_inner().ok()?;
+        Some((k.to_vec(), v.to_vec()))
+    })
+    .collect()
+}
```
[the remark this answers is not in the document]
zaidoon1: fixed

src/tree/mod.rs
```diff
@@ -840,6 +853,47 @@ impl Tree {
...
+            } else {
+                // No extractor configured: rely on full-key Bloom as usual.
+                table.get(key, seqno, key_hash)?
+            }
+        } else {
+            // Incompatible extractor or mismatch: never trust the filter.
+            table.get_without_filter(key, seqno)?
+        };
+
+        Ok(item)
```
marvin-j97: I think this is a more succint version of the same function:

```suggestion
        // Determine compatibility of table's stored extractor with current config
        let allow_filter =
            table.prefix_filter_allowed(config.prefix_extractor.as_ref().map(|e| e.name()));

        if allow_filter {
            if let Some(ex) = config.prefix_extractor.as_ref() {
                // If prefix filtering is allowed and an extractor is configured, consult the
                // prefix-aware filter first and skip on a definite negative.
                if table.maybe_contains_prefix(key, ex.as_ref())? == Some(false) {
                    return Ok(None);
                }
            } else {
                // No extractor configured: rely on full-key Bloom as usual.
                return table.get(key, seqno, key_hash);
            };
        }

        // Incompatible/no extractor or mismatch: never trust the filter.
        table.get_without_filter(key, seqno)
```

src/tree/mod.rs
```diff
@@ -840,6 +853,47 @@ impl Tree {
...
+    }
```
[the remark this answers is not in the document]
marvin-j97: This seems very wrong

src/tree/mod.rs
zaidoon1: updated

src/table/writer/filter/full.rs
```diff
@@ -49,6 +49,11 @@ impl<W: std::io::Write + std::io::Seek> FilterWriter<W> for FullFilterWriter {
         Ok(())
     }
 
+    fn register_bytes(&mut self, bytes: &[u8]) -> crate::Result<()> {
+        self.bloom_hash_buffer.push(Builder::get_hash(bytes));
+        Ok(())
+    }
```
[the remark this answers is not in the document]
zaidoon1: done

src/table/inner.rs
```diff
@@ -50,6 +50,10 @@ pub struct Inner {
     /// Pinned AMQ filter
     pub pinned_filter_block: Option<FilterBlock>,
 
+    /// Name of the prefix extractor used when this table was created, if any.
+    /// Used to determine compatibility with the current extractor for prefix-aware filtering.
+    pub prefix_extractor_name: Option<String>,
```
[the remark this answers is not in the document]
zaidoon1: fixed

src/prefix.rs
```diff
@@ -0,0 +1,364 @@
...
+///     fn name(&self) -> &str {
+///         "fixed_prefix"
+///     }
+/// }
+///
+/// let ex = FixedPrefixExtractor(3);
+/// assert_eq!(ex.name(), "fixed_prefix");
+/// assert_eq!(ex.extract_first(b"abcdef"), Some(b"abc".as_ref()));
+/// assert_eq!(ex.extract_first(b"ab"), Some(b"ab".as_ref()));
+/// ```
```
marvin-j97, line 72: We already export `FixedPrefixExtractor` as a predefined extractor anyway, so no need to repeat its implementation in the docs.

src/prefix.rs
```diff
@@ -0,0 +1,364 @@
...
+    /// An empty iterator means the key is "out of domain" and won't be added to the filter.
+    fn extract<'a>(&self, key: &'a [u8]) -> Box<dyn Iterator<Item = &'a [u8]> + 'a>;
+
+    /// Extracts the first prefix from a key.
+    ///
+    /// By default, this is derived from `extract`, meaning it is equivalent to `extract(key).next()`,
+    /// however it can be overridden to skip the Box allocation of `extract` in some cases.
+    ///
+    /// Implementations that override this method must remain semantically identical to
+    /// `self.extract(key).next()`: return `None` iff `extract` would yield no prefixes,
```
marvin-j97:
```suggestion
    /// `self.extract(key).next()`: return `None` if `extract` would yield no prefixes,
```

src/prefix.rs
zaidoon1: done

src/prefix.rs
zaidoon1: fixed

src/range.rs
```diff
@@ -165,17 +167,31 @@ impl TreeIter {
...
                                     range.end_bound().map(|x| &x.user_key).cloned(),
-                                ))
-                                .filter(move |item| match item {
-                                    Ok(item) => seqno_filter(item.key.seqno, seqno),
-                                    Err(_) => true,
-                                });
-
-                            iters.push(Box::new(reader));
+                                );
+                                if table.should_skip_range_by_prefix_filter(&tmp_range, ex.as_ref())
```
marvin-j97: I think there may be a problem with this (maybe a specialization for `tree.prefix()` is required):

When configuring a prefix extractor e.g. `FixedPrefixExtractor::new(1)`, with keys = "abc", "abd", "zde", you would expect queries over prefix = "h" to skip the table (because prefixes = "a", "z").

However, because prefixes are first converted to ranges, the extracted start and end prefixes do not match for the prefix, so `Table::should_skip_range_by_prefix_filter` always returns false.
And actually in a run reader, `common_prefix` will always be None, so using a prefix extractor for a prefix across a run is (almost) always disqualified.

I think checking that `tree.prefix(...)` benefits from extractors is currently a gap in test coverage.

```rs
#[test]
#[cfg(feature = "metrics")]
fn prefix_simple_fixed() -> lsm_tree::Result<()> {
    let folder = tempfile::tempdir()?;

    let tree = Config::new(
        &folder,
        SequenceNumberCounter::default(),
        SequenceNumberCounter::default(),
    )
    .prefix_extractor(Arc::new(FixedPrefixExtractor::new(1)))
    .open()?;

    tree.insert(b"abc", b"1", 0);
    tree.insert(b"abd", b"2", 0);
    tree.insert(b"zde", b"3", 0);
    tree.flush_active_memtable(0)?;

    {
        let before = tree.metrics().filter_queries();

        // Prefix query over "h" returns 0
        let keys: Vec<_> = tree
            .prefix(b"h", SeqNo::MAX, None)
            .map(|g| g.key().unwrap())
            .collect();
        assert_eq!(keys.len(), 0);

        let after = tree.metrics().filter_queries();
        assert!(after > before);
    }

    Ok(())
}
```

src/run_reader.rs
```diff
@@ -15,46 +18,153 @@ pub struct RunReader {
...
+            match (start_first, end_first) {
+                (Some(s), Some(e)) if s == e => Some(s),
+                _ => None,
+            }
+        } else {
+            None
+        };
+
+        // Early optimization
+        if let Some(ex) = extractor.clone() {
```
marvin-j97:
```suggestion
        if let Some(ex) = &extractor {
```

zaidoon1: fixed

src/range.rs
zaidoon1: fixed this and various other issues, let me know what you think

src/table/writer/filter/full.rs
marvin-j97, line 1: I don't actually know how RocksDB handles this - but I think filters built with extracted prefixes can be quite bloated in size: Consider a workload where we have `<key>#<version>` as user keys and we extract `<key>` using prefix extractor. If we have 10 versions on average per key, the bloom hash buffer will have `10 * key` buffered bytes, which will make the filter 10x as large as it needs to be (because in reality our number of prefixes only equals the number of keys).

zaidoon1: I took a look and RocksDB deduplicates adjacent hashes. Since keys with the same prefix are sorted adjacently in an SST file, this effectively deduplicates all prefix hashes within a data block / partition. For the example you gave: `<key>#<version>` with 10 versions per key, all 10 versions are adjacent, so only 1 hash is added.

marvin-j97: That probably does not work when a single key creates multiple prefixes though because then you have in the hash buffer:
```
-- key 1
hash of prefix 1
hash of prefix 2
-- key 2
hash of prefix 1
hash of prefix 2
-- key 3
hash of prefix 1
hash of prefix 2
```

So I guess a `sort()` step is needed before deduping (but all of this can be skipped if no prefix extractor is configured).

zaidoon1: that's correct. and Rocksdb doesn't have have to worry about it because it doesn't support multi prefix extractors. I'll work on pushing a commit to do the deduping correctly

src/table/mod.rs
```diff
@@ -310,6 +373,49 @@ impl Table {
...
+        extractor: &dyn crate::prefix::PrefixExtractor,
+    ) -> crate::Result<Option<bool>> {
+        let filter_block = self.load_filter_block_for_key(key)?;
+
+        if let Some(filter_block) = filter_block {
+            #[cfg(feature = "metrics")]
+            {
+                use std::sync::atomic::Ordering::Relaxed;
+                self.metrics.filter_queries.fetch_add(1, Relaxed);
+            }
```
marvin-j97: I think this negatively affects filter metrics - because the filter queries are incremented, but on hit there is no increment for filter hits.

marvin-j97:
I have run a couple of benchmarks with varying number of items and duration, and I've been getting mixed results so far. Even when artificially making the workload IO-bound, the performance improvements were often <10%.
In the column store workload (with a fixed prefix of 48), the performance actually degraded by 10%, but I'm not yet really sure why.

zaidoon1:
> I have run a couple of benchmarks with varying number of items and duration, and I've been getting mixed results so far. Even when artificially making the workload IO-bound, the performance improvements were often <10%. In the column store workload (with a fixed prefix of 48), the performance actually degraded by 10%, but I'm not yet really sure why.

can you give me the commands you are running so i can try to reproduce this locally and dig deeper?

marvin-j97:
> can you give me the commands you are running so i can try to reproduce this locally and dig deeper?

Feed:
```nu
let data_dir = ".data"
let seconds = 10 * 60
let cache_bytes = 256_000_000

alias bench = cargo run -r --features mimalloc,metrics --

bench run --cache-size $cache_bytes --auto-granularity 80 --lsm-use-partitioned-meta --data-dir $data_dir --seconds $seconds --out "pfx_.jsonl" --display-name "fjall 3.1.2" --backend fjall3 feed --users 5000000 --zipf-exponent 0.5
bench run --cache-size $cache_bytes --auto-granularity 80 --lsm-use-partitioned-meta --data-dir $data_dir --seconds $seconds --out "pfx_with.jsonl" --display-name "fjall 3.1.2 pfx" --lsm-prefix-extractor --color "#eb5aaaff" --backend fjall3 feed --users 5000000 --zipf-exponent 0.5
bench report pfx_*.jsonl --out report_pfx.html
google-chrome-stable report_pfx.html
```

Column Store:
```nu
let data_dir = ".data";
let seconds = 20 * 60;
let cache_bytes = 1_000_000_000;
let value_size = 100;
let db_count = 1_000;
let threads = 4;

alias bench = cargo run -r --features mimalloc,metrics --

bench run --auto-granularity 100 --cache-size $cache_bytes --lsm-block-size 16000 --lsm-use-partitioned-meta --data-dir $data_dir --seconds $seconds --out "cs_pfx_.jsonl" --display-name "fjall 3.1.2" --backend fjall3 column-store --value-size $value_size --readers $threads --database-count $db_count
RUST_LOG=warn bench run --auto-granularity 100 --cache-size $cache_bytes --lsm-block-size 16000 --data-dir $data_dir --seconds $seconds --out "cs_pfx_with.jsonl" --display-name "fjall 3.1.2 pfx" --lsm-prefix-extractor --color "#eb5aaaff" --backend fjall3 column-store --value-size $value_size --readers $threads --database-count $db_count
bench report cs_pfx_*.jsonl --out report_cs_pfx.html
google-chrome-stable report_cs_pfx.html
```

Note that I replaced `--lsm-prefix-extractor-len` with a workload-dependent extractor.

src/table/mod.rs
zaidoon1: pushed a commit to fix this, let me know what you think

src/table/mod.rs
```diff
@@ -623,6 +727,92 @@ impl Table {
...
+                    if matches!(self.probe_prefix_filter(sk, extractor), Ok(Some(false))) {
+                        #[cfg(feature = "metrics")]
+                        {
+                            use std::sync::atomic::Ordering::Relaxed;
+                            self.metrics.filter_queries.fetch_add(1, Relaxed);
+                            self.metrics.io_skipped_by_filter.fetch_add(1, Relaxed);
+                        }
+
+                        return true;
+                    }
```
marvin-j97, line 862: Need to increment the filter queries even if there is no true positive.
Otherwise the FPR will be too low.
(And everywhere else)

```suggestion
                if let Some(sk) = start_key {
                    #[cfg(feature = "metrics")]
                    self.metrics.filter_queries.fetch_add(1, Relaxed);

                    if matches!(self.probe_prefix_filter(sk, extractor), Ok(Some(false))) {
                        #[cfg(feature = "metrics")]
                        {
                            use std::sync::atomic::Ordering::Relaxed;

                            self.metrics.io_skipped_by_filter.fetch_add(1, Relaxed);
                        }

                        return true;
                    }
```

zaidoon1:
> > can you give me the commands you are running so i can try to reproduce this locally and dig deeper?
> 
> Feed:
> 
> ```nu
> let data_dir = ".data"
> let seconds = 10 * 60
> let cache_bytes = 256_000_000
> 
> alias bench = cargo run -r --features mimalloc,metrics --
> 
> bench run --cache-size $cache_bytes --auto-granularity 80 --lsm-use-partitioned-meta --data-dir $data_dir --seconds $seconds --out "pfx_.jsonl" --display-name "fjall 3.1.2" --backend fjall3 feed --users 5000000 --zipf-exponent 0.5
> bench run --cache-size $cache_bytes --auto-granularity 80 --lsm-use-partitioned-meta --data-dir $data_dir --seconds $seconds --out "pfx_with.jsonl" --display-name "fjall 3.1.2 pfx" --lsm-prefix-extractor --color "#eb5aaaff" --backend fjall3 feed --users 5000000 --zipf-exponent 0.5
> bench report pfx_*.jsonl --out report_pfx.html
> google-chrome-stable report_pfx.html
> ```
> 
> Column Store:
> 
> ```nu
> let data_dir = ".data";
> let seconds = 20 * 60;
> let cache_bytes = 1_000_000_000;
> let value_size = 100;
> let db_count = 1_000;
> let threads = 4;
> 
> alias bench = cargo run -r --features mimalloc,metrics --
> 
> bench run --auto-granularity 100 --cache-size $cache_bytes --lsm-block-size 16000 --lsm-use-partitioned-meta --data-dir $data_dir --seconds $seconds --out "cs_pfx_.jsonl" --display-name "fjall 3.1.2" --backend fjall3 column-store --value-size $value_size --readers $threads --database-count $db_count
> RUST_LOG=warn bench run --auto-granularity 100 --cache-size $cache_bytes --lsm-block-size 16000 --data-dir $data_dir --seconds $seconds --out "cs_pfx_with.jsonl" --display-name "fjall 3.1.2 pfx" --lsm-prefix-extractor --color "#eb5aaaff" --backend fjall3 column-store --value-size $value_size --readers $threads --database-count $db_count
> bench report cs_pfx_*.jsonl --out report_cs_pfx.html
> google-chrome-stable report_cs_pfx.html
> ```
> 
> Note that I replaced `--lsm-prefix-extractor-len` with a workload-dependent extractor.

I believe i found the issue. When a prefix extractor was configured, the filter only contained prefix hashes, full-key hashes were not written. For point reads where the key doesn't exist but its prefix DOES exist in the table (e.g., looking up user_alice when user_bob exists — same prefix "user"):
- Without prefix extractor: full-key Bloom catches the nonexistent key ~99% of the time → no data block read
- With prefix extractor: prefix filter says "maybe present" → get_without_filter → reads data blocks → finds nothing → wasted I/O

Rocksdb solves it using the whole_key_filtering option that is default enabled which adds both the prefix + the whole key to the bloom unless you disable whole_key_filtering in which case only the prefix is added: https://github.com/facebook/rocksdb/wiki/RocksDB-Bloom-Filter#prefix-vs-whole-key

I'll add the same thing here and make it configurable because I have workloads in prod that disables whole_key_filtering because I don't do point lookups so I avoid the wasted space.

marvin-j97:
> I believe i found the issue. When a prefix extractor was configured, the filter only contained prefix hashes, full-key hashes were not written. For point reads where the key doesn't exist but its prefix DOES exist in the table (e.g., looking up user_alice when user_bob exists — same prefix "user"):
> 
> * Without prefix extractor: full-key Bloom catches the nonexistent key ~99% of the time → no data block read
> * With prefix extractor: prefix filter says "maybe present" → get_without_filter → reads data blocks → finds nothing → wasted I/O
> 
> Rocksdb solves it using the whole_key_filtering option that is default enabled which adds both the prefix + the whole key to the bloom unless you disable whole_key_filtering in which case only the prefix is added: https://github.com/facebook/rocksdb/wiki/RocksDB-Bloom-Filter#prefix-vs-whole-key
> 
> I'll add the same thing here and make it configurable because I have workloads in prod that disables whole_key_filtering because I don't do point lookups so I avoid the wasted space.

Yeah, I'm aware of full keys. I have actually manually extracted the full keys in the custom extractor for the `Feed` workload. That should take care of it unless the extracted full key is not actually used (haven't checked tbh). However, the column store workload does not even do any point reads.

Edit: Actually manually extracting the full keys may not work so that may at least explain why I saw the feed workload being slower in point reads sometimes.

zaidoon1:
> However, the column store workload does not even do any point reads.

I'll look into this more.

marvin-j97:
So with the whole key filtering at least point reads are the same as before. The workload I've tested is not very write intensive so L0 only gets up to 4 tables shortly, so I guess for a write intensive workload, or setting L0 threshold to be higher, prefix filtering would help more.

src/table/mod.rs
zaidoon1: updated

zaidoon1:
I've pushed up a commit to fix the regression:

```
=== COLUMN STORE BENCHMARK COMPARISON (5 min each) ===
Config                          Scans/s  vs base    Lat(us)  Filter(MiB)  DiskRd(MiB)
-------------------------------------------------------------------------------------
Baseline (no prefix)             32,155    +0.0%       32.3        304.8     64,062.4
Prefix 48B (pre-opt)             28,755   -10.6%      966.9        609.7     52,248.1
Prefix 48B (optimized)           37,330   +16.1%       27.2        609.4     53,634.5
Prefix 48B (no wkf)              35,452   +10.3%       28.8        304.6     66,498.8
```

zaidoon1:
and the extract vs extract_first optimization:
```
=== FEED WORKLOAD: extract() vs extract_first() (5 min each) ===
Metric                      Box (extract)   extract_first     Change
----------------------------------------------------------------------
Point read ops/s                  245,881         290,299     +18.1%
Range scan ops/s                   63,874          72,052     +12.8%
Write ops/s                       235,287         278,494     +18.4%
Point read lat (us)                   5.0             3.5     -30.4%
Range scan lat (us)                  19.0            14.2     -25.1%
Total point reads              14,846,503      16,609,555     +11.9%
…
```

zaidoon1:
perf optimizations wise. I'm not seeing anything else to optimize so I think I can stop here. I'll wait for you to review the latests commits and then address any feedback you may have before i fix the merge conflict.

marvin-j97:
I haven't thought about it too much until now, but at this point only `extract_first` is really used, and makes me question how well multi prefixes really work. Because if you extract a multiple prefixes from a key, let's say `<company_id>#<user_id>#...` -> `["<company>#", "<company>#<user>#"]` naturally that allows to filter for prefix queries over `company` and `(company + user)`.

However: as it currently stands, for a `(company + user)` query, extract_first will only check for `company#`, and I'm unsure whether going through all prefixes warrants the heap allocation + filter lookup overhead.
Obviously `company + user` has a higher cardinality, so it would be more exact for filtering, I guess.

zaidoon1:
> I haven't thought about it too much until now, but at this point only `extract_first` is really used, and makes me question how well multi prefixes really work. Because if you extract a multiple prefixes from a key, let's say `<company_id>#<user_id>#...` -> `["<company>#", "<company>#<user>#"]` naturally that allows to filter for prefix queries over `company` and `(company + user)`.
> 
> However: as it currently stands, for a `(company + user)` query, extract_first will only check for `company#`, and I'm unsure whether going through all prefixes warrants the heap allocation + filter lookup overhead. Obviously `company + user` has a higher cardinality, so it would be more exact for filtering, I guess.

ok i've pushed three commits to address all of this:

The `extract()` iterator is still used on the write path to register ALL prefix hashes in the filter, so multi-prefix extractors work correctly for both coarse and fine-grained queries.
For reads, we now use `extract_last` (the most specific prefix) instead of `extract_first` (the coarsest) when probing the filter. This is computed once per query in `RunReader::new` and `should_skip_range_by_prefix_filter`, not per table — so the one-time `extract()` Box allocation is amortized. The per-table probe uses a precomputed hash with zero allocation.
For a multi-prefix extractor like `["company#", "company#user#"]`:
- `tree.prefix("company#user#hello")` → probes with `hash("company#user#")` (most specific stable prefix). Tables with the company but not this specific user are skipped.
- `tree.prefix("company#")` → probes with `hash("company#")` (only prefix available at this hint length). Tables without this company are skipped.

The approach adapts to the hint length automatically. A stability guard (`extract_last(hint) == extract_last(hint + "\0")`) ensures the chosen prefix is stable across the query range — if the most specific prefix isn't stable, it falls back to `extract_first`.
For single-prefix extractors, `extract_last` is overridden to delegate to `extract_first` (zero allocation, identical behavior). No overhead change.
For point reads with `whole_key_filtering=true` (the default), the prefix filter is skipped entirely — point reads go straight to the full-key Bloom which is strictly more precise. The prefix filter only activates for range/prefix scans where it provides table-level pruning.

zaidoon1:
Ran rust-storage-bench (Feed and Column-Store) comparing the basic SST-level prefix Bloom in this PR against RocksDB. The bench harness only calls `set_prefix_extractor` on RocksDB — no memtable bloom, no hash data block index, no other RocksDB-specific features — so this isolates the basic SST-level prefix Bloom comparison.

Same flags everywhere: `--lsm-use-partitioned-meta` on every config, 5-minute measurement window after pre-ingest, fresh data dir per run. M4 Pro / 48 GiB RAM, lsm-tree branch at `9cf39ad9`.

## Feed (5M users, 255M items, 11-byte user prefix)

| | write p50 | write p99 | point_read p50 | point_read p99 | range_read p50 | range_read p99 |
|---|---:|---:|---:|---:|---:|---:|
| lsm-tree baseline | 4.68 µs | 14.92 µs | 99.74 µs | 255.34 µs | 8.35 µs | 121.82 µs |
| lsm-tree +pfx-11  | **4.68 µs** | **14.92 µs** | **95.83 µs** | **213.28 µs** | **7.56 µs** | **117.05 µs** |
| rocksdb  baseline | 5.27 µs | 16.82 µs | 110.23 µs | 255.34 µs | 10.20 µs | 137.36 µs |
| rocksdb  +pfx-11  | 5.38 µs | 17.51 µs | 114.73 µs | 255.34 µs | 10.62 µs | 142.96 µs |

lsm-tree (with or without the extractor) is faster than RocksDB on every metric. Comparing best-vs-best (both with prefix), lsm-tree wins by 13–29% across the board. Enabling the prefix extractor on lsm-tree gives a real benefit: point_read p99 −16.5%, range_read p50 −9.5%.

## Column-Store (1000 DBs, 4 readers, 100B values, 48-byte prefix)

| | write p50 | write p99 | range_read p50 | range_read p90 | range_read p99 |
|---|---:|---:|---:|---:|---:|
| lsm-tree baseline | 3.61 µs | 33.87 µs | 18.22 µs | 38.19 µs | 217.59 µs |
| lsm-tree +pfx-48  | 3.20 µs | **10.20 µs** | **17.86 µs** | **51.55 µs** | 240.47 µs |
| rocksdb  baseline | **2.95 µs** | 8.69 µs | 18.96 µs | 88.46 µs | 282.20 µs |
| rocksdb  +pfx-48  | **2.89 µs** | **8.02 µs** | 18.59 µs | 75.38 µs | **240.47 µs** |

Mixed here. Range reads: lsm-tree wins p50 by 3.9% and p90 by 31.6%, p99 ties. Writes: RocksDB wins p50 by 9.7% and p99 by 21.3%.

The column-store write gap isn't a prefix-filter issue — even RocksDB baseline (no extractor) beats lsm-tree baseline on writes (p99 8.69 vs 33.87 µs) under this workload. That's a write-throttling / compaction-backpressure characteristic, not the filter code path. Adding the prefix extractor narrows lsm-tree's write tail substantially (33.87 → 10.20 µs, since smaller filters reduce compaction backlog) but doesn't close the gap entirely.

Both backends benefit from enabling the prefix extractor on column-store: RocksDB range_read p99 −14.8%, lsm-tree range_read p50 −2%.

Caveat: the rocksdb column-store runs hit `Corruption: Mismatch in unique ID on table file` mid-run — a known rust-rocksdb 0.44.2 issue on macOS (the `Actual` value `{6918025063187695999, 149668763839083335}` is a constant across runs, suggesting an uninitialised-memory read in `verify_sst_unique_id_in_manifest`). The bench produced final histograms before crashing, and the values reproduced consistently across two attempts, so I'm using them with the caveat noted. Linux runs don't hit this.

## Commands

```bash
# Feed
$BENCH run --cache-size 256000000 --auto-granularity 80 --lsm-use-partitioned-meta \
  --data-dir /tmp/feed_data --seconds 300 --out feed_lsm_baseline.jsonl \
  --display-name "lsm-tree baseline" --backend fjall3 \
  feed --users 5000000 --zipf-exponent 0.5

# repeat with --lsm-prefix-extractor-len 11 for the +pfx variant
# repeat with --backend rocksdb for the rocksdb variants

# Column-Store
$BENCH run --auto-granularity 100 --cache-size 1000000000 --lsm-block-size 16000 \
  --lsm-use-partitioned-meta --data-dir /tmp/cs_data --seconds 300 \
  --out cs_lsm_baseline.jsonl --display-name "lsm-tree cs baseline" --backend fjall3 \
  column-store --value-size 100 --readers 4 --database-count 1000

# repeat with --lsm-prefix-extractor-len 48 for +pfx
# repeat with --backend rocksdb for the rocksdb variants
```

Built with `cargo build -r --features mimalloc,rocksdb`. The `metrics` feature is broken in this fork (gates rocksdb code unconditionally), so left it off. Cleared the data dir between runs.

## Summary

Basic SST-level prefix Bloom in this PR is at parity with or ahead of RocksDB's. lsm-tree wins decisively on Feed across all metrics. On Column-Store, lsm-tree wins range reads; RocksDB wins writes — but the write difference is a write-throttling characteristic visible in the baseline too, not the prefix filter.

Issue #190, opened 2025-11-07
Expose Ingestion API with Inversion of Control


marvin-j97:
The current ingest API is too limited, it does not allow fallible inputs, and also does not allow writing tombstones.

An API more similar to a WriteBatch could be exposed:

e.g.

```rs
use lsm_tree::Ingestion;

let ingestion = Ingestion::new(tree)?;

for item in ... {
  ingestion.write(k, v)?;
}

ingestion.write_tombstone(k)?;

ingestion.finish()?;
```

This basically already exists, just needs to be exposed and tested. Then remove the old API (tree.ingest).
Caution that the blob tree is a bit more involved.

PR #191, opened 2025-11-08
Expose Ingestion API with Inversion of Control

zaidoon1:
resolve https://github.com/fjall-rs/lsm-tree/issues/190


tests/tree_bulk_ingest.rs
```diff
@@ -107,16 +123,20 @@ fn blob_tree_bulk_ingest() -> lsm_tree::Result<()> {
...
-        &seqno,
-        &visible_seqno,
-    )?;
+    let tree = match tree {
+        AnyTree::Blob(t) => t,
+        _ => unreachable!(),
+    };
+
+    let seq = seqno.next();
+    let mut ingestion = BlobIngestion::new(&tree)?.with_seqno(seq);
```
marvin-j97: Because in fjall (and in general), the type of the tree is kind of opaque (AnyTree), it would be better for the ingestion API to just accept an AnyTree, and internally handle the difference (maybe using `dyn`).

src/tree/mod.rs
```diff
@@ -742,13 +712,18 @@ impl Tree {
...
 
+        let mut best: Option<InternalValue> = None;
+
         for level in version.iter_levels() {
             for run in level.iter() {
                 // NOTE: Based on benchmarking, binary search is only worth it with ~4 tables
                 if run.len() >= 4 {
                     if let Some(table) = run.get_for_key(key) {
                         if let Some(item) = table.get(key, seqno, key_hash)? {
-                            return Ok(ignore_tombstone_value(item));
```
marvin-j97: This will make a point read cascade through all levels, even though through compaction we guarantee that the first item we find (in the "most upper" run), will have the highest version, no?

tests/tree_bulk_ingest.rs
zaidoon1: done

src/tree/mod.rs
zaidoon1: that's correct, that was my mistake and should be fixed now.

src/tree/mod.rs
```diff
@@ -84,19 +84,30 @@ impl AbstractTree for Tree {
         let version_history_lock = self.version_history.read().expect("lock is poisoned");
         let super_version = version_history_lock.get_version_for_snapshot(seqno);
 
-        if let Some(entry) = super_version.active_memtable.get(key, seqno) {
-            return Ok(ignore_tombstone_value(entry));
-        }
+        // Track the newest visible version across all sources
+        let mut best: Option<InternalValue> = super_version.active_memtable.get(key, seqno);
```
marvin-j97: Same for this actually; if we find a KV in a memtable, we want to terminate early.

zaidoon1: done

zaidoon1: actually what i did is not right as tests are failing. I need to push another fix

zaidoon1: should be good now

src/tree/mod.rs
```diff
@@ -84,19 +84,37 @@ impl AbstractTree for Tree {
...
-        if let Some(entry) = super_version.active_memtable.get(key, seqno) {
-            return Ok(ignore_tombstone_value(entry));
+        // Track best candidate by seqno across all sources
+        let mut best: Option<InternalValue> = super_version.active_memtable.get(key, seqno);
+        // If the memtable hit matches the snapshot seqno, it's the maximal visible version
+        if let Some(ref entry) = best {
+            if entry.key.seqno == seqno {
+                return Ok(ignore_tombstone_value(entry.clone()));
+            }
+        }
```
marvin-j97: I don't understand this change (and the subsequent ones) - `memtable.get` takes care of selecting the maximum version that satisfies the snapshot watermark. If we get anything out of a memtable, we want to return as early as possible always.

There is no way there can be a higher version after that because of the invariants we have:
(1) active memtables always shadow sealed memtables
(2) sealed memtables are ordered FIFO and shadow tables
(3) tables are arranged FIFO if they overlap (because flushes always push the new table on top of L0)

zaidoon1: While the source ordering invariants are correct for “normal” writes (active → sealed → tables), they don’t hold universally because the ingestion path writes directly to tables with an explicit seqno via Ingestion::with_seqno(...). 

That means we can (legally) have an older value in the active memtable (e.g., seqno=0) and a newer tombstone in tables (seqno=10). So an unconditionally early-returning on a memtable hit is incorrect in ingestion scenarios—it would return a stale value and ignore a newer table tombstone.

I've pushed a change to make things more efficient but i'm still checking various things to handle the ingestion api. However, if i'm misunderstanding what the ingestion api can legally do, I'm happy to revert these changes as well as the 3 tests I added that fail without these changes:

```
the following three tests now fail:

 PASS [   0.566s] lsm-tree::dirty_snapshot dirty_snapshot_after_drop_range
        FAIL [   0.368s] lsm-tree::ingestion_api blob_ingestion_only_tombstones_does_not_create_blob_files
  stdout ───

    running 1 test
    test blob_ingestion_only_tombstones_does_not_create_blob_files ... FAILED

    failures:

    failures:
        blob_ingestion_only_tombstones_does_not_create_blob_files

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 11 filtered out; finished in 0.36s
    
  stderr ───

    thread 'blob_ingestion_only_tombstones_does_not_create_blob_files' (37376406) panicked at tests/ingestion_api.rs:241:9:
    assertion failed: tree.get(key.as_bytes(), SeqNo::MAX)?.is_none()
    note: run with RUST_BACKTRACE=1 environment variable to display a backtrace

        FAIL [   0.361s] lsm-tree::ingestion_api blob_tree_ingestion_tombstones_delete_existing_keys
  stdout ───

    running 1 test
    test blob_tree_ingestion_tombstones_delete_existing_keys ... FAILED

    failures:

    failures:
        blob_tree_ingestion_tombstones_delete_existing_keys

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 11 filtered out; finished in 0.35s
    
  stderr ───

    thread 'blob_tree_ingestion_tombstones_delete_existing_keys' (37376473) panicked at tests/ingestion_api.rs:193:9:
    assertion failed: tree.get(key.as_bytes(), SeqNo::MAX)?.is_none()
    note: run with RUST_BACKTRACE=1 environment variable to display a backtrace

   
        FAIL [   0.386s] lsm-tree::ingestion_api tree_ingestion_tombstones_delete_existing_keys
  stdout ───

    running 1 test
    test tree_ingestion_tombstones_delete_existing_keys ... FAILED

    failures:

    failures:
        tree_ingestion_tombstones_delete_existing_keys

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 11 filtered out; finished in 0.37s
    
  stderr ───

    thread 'tree_ingestion_tombstones_delete_existing_keys' (37376529) panicked at tests/ingestion_api.rs:21:9:
    assertion failed: tree.get(key.as_bytes(), SeqNo::MAX)?.is_none()
    note: run with RUST_BACKTRACE=1 environment variable to display a backtrace

────────────
     Summary [  27.092s] 358 tests run: 355 passed (1 leaky), 3 failed, 5 skipped
        FAIL [   0.368s] lsm-tree::ingestion_api blob_ingestion_only_tombstones_does_not_create_blob_files
        FAIL [   0.361s] lsm-tree::ingestion_api blob_tree_ingestion_tombstones_delete_existing_keys
        FAIL [   0.386s] lsm-tree::ingestion_api tree_ingestion_tombstones_delete_existing_keys
error: test run failed
```

marvin-j97: Right. I think I temporarily removed the comments, but basically it boils down to this:

- Ingesting when there is already a non-empty memtable is not supported
- To support it, the memtable would need to be flushed first (not a necessity for this PR)
- While an ingestion is happening, no writes are allowed to go into the memtable (or other ingestions for that matter) - I used to have a write lock on the active memtable, but now with the coarse lock around SuperVersion that's probably a bit harder to do without blocking readers
- Ultimately, the invariants of how item versions are organized needs to stay, and thus the read path can stay as it was before

zaidoon1: I've restored the original logic. I've kept 
```
// Avoid holding the read lock across potential I/O in table lookups
drop(version_history_lock);` 
```

as it seems like it's good to do. 

I've also updated the code to prevent writes while ingestion is happening and flushing memtables, etc... before starting ingestion. If you don't think my implementation is good, I can remove all that.

src/tree/ingest.rs
```diff
@@ -28,6 +31,45 @@ impl<'a> Ingestion<'a> {
...
+        // Block concurrent writes and concurrent ingestions for the duration of this ingestion.
+        let ingestion_guard = tree.0.ingestion_lock.write().expect("lock is poisoned");
+
+        // If the active memtable has data, flush it first to preserve strict recency invariants
+        // for the read path (active > sealed > tables). This avoids overlapping in-memory state
+        // with newly ingested on-disk tables.
+        tree.flush_active_memtable(SeqNo::MAX)?;
+
+        // Flush any remaining sealed memtables synchronously into tables, then register them.
+        if tree.sealed_memtable_count() > 0 {
```
marvin-j97: In Fjall, these sealed memtables are queued for flushing. In this situation we are basically preempting the flush worker and stealing its work. So it's important fjall can handle this to avoid double flushing.
I need to check that.

src/tree/ingest.rs
```diff
@@ -28,6 +31,45 @@ impl<'a> Ingestion<'a> {
     ///
     /// Will return `Err` if an IO error occurs.
     pub fn new(tree: &'a Tree) -> crate::Result<Self> {
+        // Block concurrent writes and concurrent ingestions for the duration of this ingestion.
+        let ingestion_guard = tree.0.ingestion_lock.write().expect("lock is poisoned");
+
+        // If the active memtable has data, flush it first to preserve strict recency invariants
+        // for the read path (active > sealed > tables). This avoids overlapping in-memory state
+        // with newly ingested on-disk tables.
+        tree.flush_active_memtable(SeqNo::MAX)?;
```
marvin-j97: We need to flush the sealed memtables in FIFO order before the active memtable.

Should be fairly easy to unit test by sealing a memtable and then, after ingestion, making sure its table does not shadow the contents of the active one.

src/tree/ingest.rs
```diff
@@ -28,6 +31,45 @@ impl<'a> Ingestion<'a> {
...
+                    tables.push(table);
+                }
+            }
+            if !tables.is_empty() {
+                tree.register_tables(&tables, None, None)?;
+            }
+        }
+
+        // Ensure invariants are satisfied before proceeding with ingestion.
+        if tree.sealed_memtable_count() > 0 {
```
marvin-j97: Checking the sealed memtable count again after flushing them screams TOCTOU, so maybe just read lock the SuperVersion lock at the top of the function? (To prevent intermittent flushes)

marvin-j97:
(Forgot to submit comments)

But as I said, I am currently changing the Flush API which should make the ingestion API much simpler.

src/tree/ingest.rs
```diff
@@ -28,6 +31,45 @@ impl<'a> Ingestion<'a> {
     ///
     /// Will return `Err` if an IO error occurs.
     pub fn new(tree: &'a Tree) -> crate::Result<Self> {
+        // Block concurrent writes and concurrent ingestions for the duration of this ingestion.
+        let ingestion_guard = tree.0.ingestion_lock.write().expect("lock is poisoned");
```
marvin-j97: With the new ingestion API, I think we don't actually need this lock anymore!
Because memtable rotations and setting up flushes block out writes, once a flush process starts, we can continue writing into the active memtable because it cannot sneak past the ongoing flush.

zaidoon1: I'm going to assume I should wait until your WIP pr lands before i update this pr?

marvin-j97: You can probably pull from the PR at this point, I think it's done

marvin-j97: The revamped flush API is merged now into main

src/tree/ingest.rs
```diff
@@ -190,14 +248,17 @@ impl<'a> Ingestion<'a> {
                     self.tree.id,
                     self.tree.config.cache.clone(),
                     self.tree.config.descriptor_table.clone(),
-                    false,
-                    false,
```
marvin-j97, line 194: I forgot to document this - but I purposefully avoided using the config for pinning because when you ingest a very large database it is assigned to L1 and will pin all those blocks which could provoke OOM.

src/error.rs
```diff
@@ -20,6 +20,9 @@ pub enum Error {
     /// Some required files could not be recovered from disk
     Unrecoverable,
 
+    /// Ingestion could not start because required preconditions were not met
+    IngestionPreconditionFailed,
```
marvin-j97: Deprecated

src/tree/ingest.rs
zaidoon1: updated

tests/tree_bulk_ingest.rs
```diff
@@ -146,15 +147,15 @@ fn blob_tree_copy() -> lsm_tree::Result<()> {
...
-        (0..ITEM_COUNT as u64).map(|x| {
-            let k = x.to_be_bytes();
-            let v = nanoid::nanoid!();
-            (k.into(), v.into())
-        }),
-        &seqno,
-        &visible_seqno,
-    )?;
+    let seq = seqno.next();
+    let mut ingestion = src.ingestion()?.with_seqno(seq);
```
marvin-j97: Theoretically if we take a sequence number x, we could have a race condition here where an insert is performed before the ingestion is setup.

Think:
```rs
let ingestion_seqno = next() = 1;

insert(next() = 2);

// ingestion now starts
  -> active memtable is flushed with the seqno=2 KV
  -> we finish the ingestion with seqno=1

the seqno=2 KV is now shadowed by seqno=1 -> invariant broken
```

It is extremely unlikely, but I think it's a possibility. I think that's why I initially let the ingestion take a seqno _after_ getting the version write lock.

zaidoon1: As far as I understand, Internally, the tree uses MVCC with sequence numbers so:

- Reads pick the highest seqno < snapshot_seqno for a given key.
- The pre-ingestion flush in Ingestion::new uses SeqNo::MAX as the GC threshold, so it does not drop newer versions.

That means even in the scenario:

- insert at seq = 2,

- ingestion run at seq = 1,

a read at SeqNo::MAX still sees the version with seq = 2, not 1.

So there isn’t an actual bug where a lower seq overwrites a higher one.

In any case, i've updated the tests just to be safe.

marvin-j97:
```rs
#[test]
fn ingestion_seqno_race_condition() -> lsm_tree::Result<()> {
    let folder = tempfile::tempdir()?;

    let seqno = SequenceNumberCounter::default();
    let tree = Config::new(&folder, seqno.clone()).open()?;

    let ingest = tree.ingestion()?;
    let mut ingest = ingest.with_seqno(seqno.next());

    {
        // Imagine another thread now writing before we get to call finish
        tree.insert("a", "a", seqno.next());
        assert_eq!(b"a", &*tree.get("a", seqno.get())?.unwrap());
        let flush_lock = tree.get_flush_lock();
        tree.rotate_memtable();
        tree.flush(&flush_lock, 0)?;
    }

    ingest.write("a".into(), "b".into())?;
    ingest.finish()?;

    assert_eq!(b"a", &*tree.get("a", seqno.get())?.unwrap());

    Ok(())
}
```
This reproduces what I meant. If we start an ingestion, but e.g. are scheduled out by the OS, we have seqno=x.
Another thread can now actually acquire seqno=(x+1), and write + flush a table.
Now we get to finish our ingestion. Because we got the seqno "too early", we will shadow the seqno=(x+1) item.

This is... very unlikely though. But it probably just shouldn't be possible.

src/tree/ingest.rs
```diff
@@ -28,6 +32,12 @@ impl<'a> Ingestion<'a> {
     ///
     /// Will return `Err` if an IO error occurs.
     pub fn new(tree: &'a Tree) -> crate::Result<Self> {
+        // Use the shared flush helper so ingestion participates in the same
+        // path as normal writes: any dirty memtable content is moved into
+        // tables before building new tables from the ingestion stream.
+        // This keeps the lookup path ordered as active > sealed > tables.
+        tree.flush_active_memtable(SeqNo::MAX)?;
```
marvin-j97:
```rs
// See ingest_dirty_snapshot.rs

#[test]
fn ingestion_dirty_snapshot() -> lsm_tree::Result<()> {
    let folder = tempfile::tempdir()?;

    let seqno = SequenceNumberCounter::default();
    let tree = Config::new(&folder, seqno.clone()).open()?;

    tree.insert("a", "a", seqno.next());
    tree.insert("a", "b", seqno.next());

    let snapshot_seqno = 1;
    assert_eq!(b"a", &*tree.get("a", snapshot_seqno)?.unwrap());

    let mut ingest = tree.ingestion()?.with_seqno(seqno.next());
    ingest.write("b".into(), "b".into())?;
    ingest.finish()?;

    assert_eq!(b"a", &*tree.get("a", snapshot_seqno)?.unwrap());

    Ok(())
}
```

Using `SeqNo::MAX` breaks snapshot reads.

The more secure solution here would be to just use watermark = 0.

tests/tree_bulk_ingest.rs
marvin-j97: This may not actually be easily solvable without either:

- locking writes as before (ugh),
- writing items with seqno=0 and assigning a global_seqno to the table files when committing the ingestion (while holding a lock), which is added on during reads - RocksDB actually does this: https://github.com/facebook/rocksdb/wiki/creating-and-ingesting-sst-files

marvin-j97: My idea would be to allow to pass in a `base_seqno` into register_tables, such that `Ingestion::finish` can take a lock, get the seqno, register, and then release. That should make the ingestion atomic and not break invariants.
Each table's seqno would be stored in the `Version` next to its full file checksum.

Table reads then need to be handled appropriately, so adding base_seqno to all `InternalValue`s that are returned.

marvin-j97: I have added global_seqno to tables in the `table-global-seqno` branch.

zaidoon1: ok back to this, I see you've made various changes and I'm catching up. What would you like me to work on to close this out?

marvin-j97: In the `table-global-seqno` branch there are the two todo!()'s left. Basically we need to move the flush into the ingestion finish, such that no writes can interfere while we flush, then take the next seqno and register with global seqno = that seqno.

zaidoon1: ok so want me to branch of that branch and make a PR against it and leave this one as is?

marvin-j97: I think you can just pull it into this PR

src/tree/ingest.rs
```diff
@@ -251,11 +272,38 @@ impl<'a> Ingestion<'a> {
...
+            let mut copy = current.clone();
+            copy.version = copy.version.with_new_l0_run(&created_tables, None, None);
+            copy
+        };
+
+        next_version.seqno = global_seqno;
+
+        // Persist the new version to disk and append it to the version history.
+        persist_version(&self.tree.config.path, &next_version.version)?;
+        version_lock.append_version(next_version);
```
marvin-j97: If `upgrade_version` gets its seqno as parameter (instead of the counter itself), maybe it could be reused instead?

zaidoon1: updated

marvin-j97 (approved):

PR #218, opened 2025-12-09
fix some clippy warnings

imor:


src/table/filter/standard_bloom/mod.rs
```diff
@@ -59,11 +59,27 @@ impl<'a> StandardBloomFilterReader<'a> {
         let hash_type = reader.read_u8()?;
         assert_eq!(0, hash_type, "Invalid bloom hash type");
 
+        #[expect(
+            clippy::cast_possible_truncation,
+            reason = "truncation is not expected to happen"
```
marvin-j97: A lot of these reasonings are basically stating the obvious, but, for example, in this case, really the reasoning is: `there cannot be a filter that is 4 GiB in size` (on 32 bit targets) - filters of a single table never even approach multiple megabytes in size typically.

src/blob_tree/gc.rs
```diff
@@ -137,6 +137,7 @@ impl ExpiredKvCallback for FragmentationMap {
         if kv.key.value_type.is_indirection() {
             let mut reader = &kv.value[..];
 
+            #[expect(clippy::expect_used, reason = "lock is expected to not be poisoned")]
```
marvin-j97: copy paste error

src/blob_tree/mod.rs
```diff
@@ -594,6 +603,7 @@ impl AbstractTree for BlobTree {
     fn get<K: AsRef<[u8]>>(&self, key: K, seqno: SeqNo) -> crate::Result<Option<crate::UserValue>> {
         let key = key.as_ref();
 
+        #[expect(clippy::expect_used, reason = "option is expected to exist")]
```
marvin-j97: copy paste error

PR #225, opened 2025-12-15
Compaction filters

### LINKED ISSUE #159: Compaction filters
**Is your feature request related to a problem? Please describe.**
Sometimes it is desired to have expiration for items in the database. For example, a session store where sessions expire, or a cache with simple expiry times. In such cases, deletion is not required to happen immediately, but it is still preferred to eventually get rid of the expired or stale data at some point.

**Describe the solution you'd like**
RocksDB allows the user to specify a [compaction filter](https://github.com/facebook/rocksdb/wiki/Compaction-Filter), but I am not sure if this idea would work in Fjall. The compaction filter can exclude k/v pairs during compaction, therefore deleting them. 

Alternatively something else that can efficiently accomplish the same.

**Describe alternatives you've considered**
- Delete expired values when reading: works (and would be required even if such a feature were implemented) but requires a value to be read before it can be deleted. Stale values will still get compacted, which might be a waste of resources.
- Periodically scan the keyspace and delete expired values: requires scanning entire keyspace, probably less efficient than compaction filter.

**Additional context**
N/A

iczero:
- Introduce `CompactionFilter` trait which can elect to drop a kv or modify its value
- Introduce `ItemAccessor` for `CompactionFilter` which can access key/value within the compaction filter
- Rename `ExpiredKvCallback` to `DroppedKvCallback` since it's not really exclusively for "expired" anymore
- Add filter support to `CompactionStream`

Compaction filters replace dropped KVs with a tombstone. If we are compacting to the last level, the tombstone will subsequently be removed by the `CompactionStream`. If the value is changed, it is replaced in place or written out to a new blob file as necessary.

Closes #159

### DIFF
--- a/src/compaction/stream.rs
+++ b/src/compaction/stream.rs
@@ -80,7 +80,7 @@ impl<I: Iterator<Item = Item>> CompactionStream<'_, I, NoFilter> {
 }
 
 impl<'a, I: Iterator<Item = Item>, F: StreamFilter + 'a> CompactionStream<'a, I, F> {
-    /// Installs a filter into this stream
+    /// Installs a filter into this stream.
     pub fn with_filter<NF: StreamFilter>(self, filter: NF) -> CompactionStream<'a, I, NF> {
         CompactionStream {
             inner: self.inner,


src/compaction/filter.rs
```diff
@@ -0,0 +1,118 @@
...
+    BlobIndirection, InternalValue, Slice,
+};
+
+/// Verdict returned by a [`CompactionFilter`].
+pub enum FilterVerdict {
+    /// Keep the item.
+    Keep,
+    /// Delete the item.
+    Drop,
+}
```
marvin-j97: RocksDB also allows transforming a value, that should be easy enough to add as well

src/compaction/worker.rs
```diff
@@ -388,14 +398,33 @@ fn merge_tables(
...
         .zero_seqnos(false);
 
+    let blobs_folder = opts.config.path.join(BLOBS_FOLDER);
+
+    // grab the filter so we can use it mutably
+    #[expect(clippy::expect_used, reason = "lock is expected to not be poisoned")]
+    let mut compaction_filter = {
+        let mut filter_lock = opts.filter.lock().expect("lock is poisoned");
+        filter_lock.take()
+    };
```
marvin-j97: Why not just construct the filter here? - then it can be fully removed from the options struct.

src/compaction/worker.rs
```diff
@@ -482,13 +515,19 @@ fn merge_tables(
 
     log::trace!("Blob fragmentation diff: {blob_frag_map:#?}");
 
+    let extra_blob_files = filter_blob_writer
+        .map(BlobFileWriter::finish)
+        .transpose()?
+        .unwrap_or(Vec::new());
```
marvin-j97: `unwrap_or_default`

src/compaction/filter.rs
```diff
@@ -0,0 +1,211 @@
...
+    vlog::{Accessor, BlobFileWriter, ValueHandle},
+    BlobIndirection, InternalValue, KvSeparationOptions, UserKey, UserValue, ValueType,
+};
+
+/// Verdict returned by a [`CompactionFilter`].
+pub enum FilterVerdict {
+    /// Keep the item.
+    Keep,
+    /// Delete the item.
+    Drop,
```
marvin-j97: RocksDB also supports replacing the value with a SingleDelete (or weak tombstone as it's called here) - it's called `kPurge` there, not sure what naming would be best... DropWeak...?

iczero: I've renamed them to `Remove` and `RemoveWeak` to better match the tree APIs, hopefully that will be a bit less confusing

marvin-j97:
I've been thinking of calling this "compaction mappers" instead of "compaction filters" because "filters" is already quite a loaded term (AMQ filters and filter() in Rust), while this operation is more akin to a filter_map really.

src/compaction/filter.rs
```diff
@@ -0,0 +1,220 @@
...
+    vlog::{Accessor, BlobFileWriter, ValueHandle},
+    BlobIndirection, InternalValue, KvSeparationOptions, UserKey, UserValue, ValueType,
+};
+
+/// Verdict returned by a [`CompactionFilter`].
+pub enum FilterVerdict {
+    /// Keep the item.
+    Keep,
+    /// Remove the item.
+    Remove,
```
marvin-j97, line 26: For very specific workloads, having a tombstone-less Drop (Destroy, ...?) verdict may also be desirable.

src/compaction/worker.rs
```diff
@@ -396,7 +396,7 @@ fn merge_tables(
         .config
         .compaction_filter_factory
         .as_ref()
-        .map(|f| f.make_filter());
+        .map(|f| f.make_filter(CompactionFilterContext { is_last_level }));
```
marvin-j97: I guess we could just give `make_filter` a reference and also pass the context reference to the compaction filters directly (`filter_item`).

marvin-j97:
> Regarding `Verdict::Destroy`, currently the compaction filter is only called for non-tombstones. Is there a use case to drop tombstones using the compaction filter?

I'm not sure. Probably not. `Destroy` is mostly something that would desirable if you use your own tombstone logic on top of compaction filters, instead of using native tombstones.

src/compaction/worker.rs
iczero: I intended something like this:
```rs
impl CompactionFilterFactory for SomeFactory {
  fn make_filter(&self, ctx: CompactionFilterContext) -> Box<dyn CompactionFilter> {
    Box::new(SomeFilter { ctx, something_else: true })
  }
}

impl CompactionFilter for SomeFilter {
  fn filter_item(&mut self, item: ItemAccessor) -> Verdict {
    if self.ctx.is_last_level {
      Verdict::Destroy
    } else {
      do_something_else
    }
  }
}
```

Passing directly to `filter_item` works but might have some lifetime variance issues due to the `&mut self` being invariant but I'm not sure. You might need lifetimes on the impl. This seems cleaner to me though.

marvin-j97: I just tried and the changes are trivial (no lifetime changes needed).

I think semantically it makes more sense to pass the context by reference. And this way there is no need to put the context into the Filter struct manually.

```rs
diff --git a/src/compaction/filter.rs b/src/compaction/filter.rs
index 44a074d0..c8194dae 100644
--- a/src/compaction/filter.rs
+++ b/src/compaction/filter.rs
@@ -57,7 +57,11 @@ pub trait CompactionFilter: Send {
     ///
     /// Returning an error will abort the running compaction.
     /// This should only be done when **strictly** necessary, such as when fetching a value fails.
-    fn filter_item(&mut self, item: ItemAccessor<'_>) -> crate::Result<Verdict>;
+    fn filter_item(
+        &mut self,
+        item: ItemAccessor<'_>,
+        ctx: &CompactionFilterContext,
+    ) -> crate::Result<Verdict>;
 
     /// Called when compaction is finished.
     fn finish(self: Box<Self>) {}
@@ -74,7 +78,7 @@ pub struct CompactionFilterContext {
 /// Trait that creates compaction filter objects for each compaction
 pub trait CompactionFilterFactory: Send + Sync + RefUnwindSafe {
     /// Returns a new compaction filter.
-    fn make_filter(&self, context: CompactionFilterContext) -> Box<dyn CompactionFilter>;
+    fn make_filter(&self, context: &CompactionFilterContext) -> Box<dyn CompactionFilter>;
 }
 
 struct AccessorShared<'a> {
@@ -163,6 +167,7 @@ pub(crate) struct StreamFilterAdapter<'a, 'b: 'a> {
     shared: AccessorShared<'a>,
     blob_opts: Option<&'a KvSeparationOptions>,
     blob_writer: &'a mut Option<BlobFileWriter>,
+    ctx: &'a CompactionFilterContext,
 }
 
 impl<'a, 'b: 'a> StreamFilterAdapter<'a, 'b> {
@@ -172,6 +177,7 @@ impl<'a, 'b: 'a> StreamFilterAdapter<'a, 'b> {
         version: &'a Version,
         blobs_folder: &'a Path,
         blob_writer: &'a mut Option<BlobFileWriter>,
+        ctx: &'a CompactionFilterContext,
     ) -> Self {
         Self {
             filter,
@@ -182,6 +188,7 @@ impl<'a, 'b: 'a> StreamFilterAdapter<'a, 'b> {
             },
             blob_opts: opts.config.kv_separation_opts.as_ref(),
             blob_writer,
+            ctx,
         }
     }
 
@@ -238,10 +245,13 @@ impl<'a, 'b: 'a> StreamFilter for StreamFilterAdapter<'a, 'b> {
             return Ok(StreamFilterVerdict::Keep);
         };
 
-        match filter.filter_item(ItemAccessor {
-            item,
-            shared: &self.shared,
-        })? {
+        match filter.filter_item(
+            ItemAccessor {
+                item,
+                shared: &self.shared,
+            },
+            &self.ctx,
+        )? {
             Verdict::Destroy => Ok(StreamFilterVerdict::Drop),
             Verdict::Keep => Ok(StreamFilterVerdict::Keep),
             Verdict::Remove => Ok(StreamFilterVerdict::Replace((
diff --git a/src/compaction/worker.rs b/src/compaction/worker.rs
index e15444da..354537c8 100644
--- a/src/compaction/worker.rs
+++ b/src/compaction/worker.rs
@@ -391,12 +391,14 @@ fn merge_tables(
 
     let blobs_folder = opts.config.path.join(BLOBS_FOLDER);
 
+    let filter_ctx = CompactionFilterContext { is_last_level };
+
     // construct the compaction filter
     let mut compaction_filter = opts
         .config
         .compaction_filter_factory
         .as_ref()
-        .map(|f| f.make_filter(CompactionFilterContext { is_last_level }));
+        .map(|f| f.make_filter(&filter_ctx));
 
     // this is used by the compaction filter if it wants to write new blobs
     let mut filter_blob_writer = None;
@@ -406,6 +408,7 @@ fn merge_tables(
         &current_super_version.version,
         &blobs_folder,
         &mut filter_blob_writer,
+        &filter_ctx,
     ));
 
     let table_writer =

```

iczero: I see, looks good to me. Will you push this change?

marvin-j97:
> Added new non_exhaustive `CompactionFilterContext` struct used in `make_filter`.
> 
> RocksDB uses the following:
> 
> ```c++
> CompactionFilter::Context context;
> context.is_full_compaction = is_full_compaction_;
> context.is_manual_compaction = is_manual_compaction_;
> context.input_start_level = start_level_;
> context.is_bottommost_level = bottommost_level_;
> context.column_family_id = cfd_->GetID();
> context.reason = TableFileCreationReason::kCompaction;
> context.input_table_properties = GetInputTableProperties();
> ```
> 
> Maybe can add the others later if wanted?

We don't actually have most of this information, so yeah it's fine as it is currently.

marvin-j97 (approved):

src/compaction/filter.rs
```diff
@@ -0,0 +1,267 @@
...
+                blobs_folder,
+            },
+            blob_opts: opts.config.kv_separation_opts.as_ref(),
+            blob_writer,
+            ctx,
+        }
+    }
+
+    /// Redirects a write to a blob file if KV separation is enabled and
+    /// the value meets the separation threshold.
```
marvin-j97, line 201: I just noticed now:

All the blobs that are written by the compaction filter end up in separate blob files. However, the compaction stream is ordered, so we can just pipe the changed value into the blob writer of the compaction stream, instead of creating our own blob file writer. Then, all the extra_blob_files logic can be removed.
However currently this is problematic because the relocating compaction stream is not really designed to accept new blobs ad-hoc (it can only pipe through old possibly compressed blobs), and the standard compaction stream is fully missing a blob writer.

So this is... fine; it _should_ be changable in the future without breaking change.

PR #229, opened 2026-01-04
New skip list

marvin-j97:
Reopening of https://github.com/fjall-rs/lsm-tree/pull/131

Stuff to do:

0. Make it compile
1. Make buffer size configurable instead of being a const generic parameter
2. Insertions of duplicates need to be rejected (not panic), to make batch writes with duplicate keys work
3. Evaluate memory overhead vs crossbeam skiplist
4. Evaluate read and write latency vs crossbeam skiplist
5. Try in a full system (heavily cached) benchmark
6. Remove crossbeam-skiplist
(7. Model test against crossbeam-skiplist(

### DIFF
--- a/src/memtable/skiplist/skipmap.rs
+++ b/src/memtable/skiplist/skipmap.rs
@@ -1,0 +1,856 @@
+// Copyright (c) 2024-present, fjall-rs
+// This source code is licensed under both the Apache 2.0 and MIT License
+// (found in the LICENSE-* files in the repository)
+
+#![allow(unsafe_code)]
+
+use equivalent::Comparable;
+
+use super::arena::Arenas;
+use std::{
+    alloc::Layout,
+    borrow::Borrow,
+    hash::Hash,
+    marker::PhantomData,
+    mem::{offset_of, ManuallyDrop},
+    ops::{Bound, RangeBounds},
+    sync::{
+        atomic::{AtomicPtr, AtomicU32, AtomicUsize, Ordering},
+        LazyLock,
+    },
+};
+
+/// A `SkipMap` is a concurrent, ordered map like a `BTreeMap`
+/// but it allows for concurrent reads and writes.
+///
+/// A tradeoff is that it does not allow for updates or deletions.
+pub struct SkipMap<K, V> {
+    arena: ArenasAllocator<K, V>,
+
+    head: BoundaryNode<K, V>,
+    tail: BoundaryNode<K, V>,
+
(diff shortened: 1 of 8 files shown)


src/memtable/skiplist/skipmap.rs
```diff
@@ -0,0 +1,856 @@
...
+        num ^= num << 5;
+        self.seed.store(num, Ordering::Relaxed);
+
+        let mut height = 1;
+        for &p in PROBABILITIES.iter() {
+            if num > p {
+                break;
+            }
+            height += 1;
+        }
```
matthew-mcallister, line 272: There is a quicker way to get an exponential distribution if you are OK with base 2 instead of base e.
```rust
const MASK: u32 = !((1 << (MAX_HEIGHT as u32 - 1)) - 1);
let x = random_u32 | MASK;
let height = x.trailing_zeros() + 1;
```

src/memtable/skiplist/skipmap.rs
```diff
@@ -0,0 +1,856 @@
...
+// This implementation is somewhat unfortunate in that it's going to
+// bounce around the SkipMap in sorted order.
+//
+// TODO: Perhaps a better design would be to keep nodes densely in
+// the arenas so that it was possible to iterate through the initialized
+// nodes without needing to traverse the links when dropping for better
+// memory locality. A downside there is that we'd need to keep fixed-sized
+// nodes. Perhaps a reasonable solution there might be to have only towers
+// taller than 1 out-of-line and then we could iterate all the nodes more
+// cheaply.
```
matthew-mcallister, line 388: This might not be worth addressing since user range queries already bounce around the skipmap in sorted order anyways...

... but, for what it's worth, you could solve this problem with minimal overhead by storing the height on each node in a clever way. Every node is aligned to 8 bytes, so the least significant 3 bits of each pointer in the tower are essentially free space. There are a few ways you could encode the height in those bits. Then you can iterate over the arena in linear order.

marvin-j97:
> I would be happy to contribute here if you want to share what you had in mind for benchmarking in points (3)-(5). Assuming you'll permit me to circle back to custom comparison functions after this PR is merged.

Benchmarking memory overhead should be pretty straightforward: just load tons of small values to see what the relative overhead of towers etc. is.

Benchmarking simple operations should be doable using criterion.

And full system benchmarks can be done using https://github.com/marvin-j97/rust-storage-bench/tree/v1, especially using the YCSB C workload.

PR #233, opened 2026-01-10
Add Guard::into_inner_if_some

asonix:
Allows passing the Key slice to the predicate and receiving a parsed value, returning that parsed value as the first element of the returned tuple. This can avoid double-parsing of keys that might happen with just Guard::into_inner_if

### DIFF
--- a/src/blob_tree/mod.rs
+++ b/src/blob_tree/mod.rs
@@ -61,7 +61,7 @@ impl IterGuard for Guard {
     fn into_inner_if_some<T>(
         self,
         pred: impl Fn(&UserKey) -> Option<T>,
-    ) -> crate::Result<Option<(T, UserValue)>> {
+    ) -> crate::Result<Result<(T, UserValue), UserKey>> {
         let kv = self.kv?;
 
         if let Some(t) = pred(&kv.key.user_key) {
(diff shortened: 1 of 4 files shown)


src/blob_tree/mod.rs
```diff
@@ -58,6 +58,27 @@ impl IterGuard for Guard {
...
     }
 
+    fn into_inner_if_some<T>(
+        self,
+        pred: impl Fn(&UserKey) -> Option<T>,
+    ) -> crate::Result<Option<(T, UserValue)>> {
+        let kv = self.kv?;
+
+        if let Some(t) = pred(&kv.key.user_key) {
+            resolve_value_handle(
```
asonix, line 68: resolve_value_handle needs access to the key slice, so the predicate can't take ownership of it.

marvin-j97:
```rs
    ) -> crate::Result<(K, Option<UserValue>)> {
```
would probably be the more flexible API as it allows reusing the parsed key even if a value was chosen not to be loaded.

asonix:
We could return a `Result<Result<(K, UserValue), UserKey)>>` maybe, so if the K can't parse the unparsed key can still be accessed

I also think we don't care about this function returning a key without loading the value. If a user wants to do that they can simply call `guard.key()` and then parse that key.

PR #234, opened 2026-01-11
Feature: Allow bypassing File Descriptor cache

### LINKED ISSUE #195: Allow fully bypassing file descriptor cache
Basically like RocksDB's open file limit = -1.
In that case, any table file is just opened and the read path can fully skip the FD cache.

```rs
let tree = C.descriptor_table(None).open()?;
// ...
```

kunalmohan:
Fix #195

This is done by introducing a `DescriptorTable::disabled()` api and making the inner field optional. The insert operations on DesciptorTable are no-op in disabled mode and get always return `None`. This looked cleaner over keeping and handling `Option<Arc<DescriptorTable>>` everywhere in the code.

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -320,9 +320,11 @@ impl Config {
         self
     }
 
+    /// Sets the file descriptor cache.
+    ///
+    /// Can be shared across trees.
     #[must_use]
-    #[doc(hidden)]
-    pub fn descriptor_table(mut self, descriptor_table: Option<Arc<DescriptorTable>>) -> Self {
+    pub fn use_descriptor_table(mut self, descriptor_table: Option<Arc<DescriptorTable>>) -> Self {
         self.descriptor_table = descriptor_table;
         self
     }
--- a/tests/blob_pinned_fd.rs
+++ b/tests/blob_pinned_fd.rs
@@ -14,7 +14,7 @@ fn blob_tree_with_pinned_fd() -> lsm_tree::Result<()> {
             SequenceNumberCounter::default(),
             SequenceNumberCounter::default(),
         )
-        .descriptor_table(None)
+        .use_descriptor_table(None)
         .with_kv_separation(Some(
             KvSeparationOptions::default().compression(lsm_tree::CompressionType::None),
         ))
@@ -40,7 +40,7 @@ fn blob_tree_with_pinned_fd() -> lsm_tree::Result<()> {
             SequenceNumberCounter::default(),
             SequenceNumberCounter::default(),
         )
-        .descriptor_table(None)
+        .use_descriptor_table(None)
         .with_kv_separation(Some(Default::default()))
         .open()?;
 


src/table/util.rs
```diff
@@ -32,6 +32,7 @@ pub struct SliceIndexes(pub usize, pub usize);
 pub fn load_block(
     table_id: GlobalTableId,
     path: &Path,
+    pinned_file_descriptor: Option<&Arc<File>>,
     descriptor_table: &DescriptorTable,
```
marvin-j97: I think it would be better for this to be unified to be a single enum.

We only have 2 cases really: we give it a descriptor table or a raw File, e.g.:

```rs
enum FileAccessor<'a> {
  File(&'a File),
  DescriptorTable(&'a DescriptorTable),
}
```

src/descriptor_table.rs
```diff
@@ -16,7 +16,7 @@ struct CacheKey(u8, u64, u64);
 
 /// Caches file descriptors to tables and blob files
 pub struct DescriptorTable {
-    inner: QuickCache<CacheKey, Item, UnitWeighter, rustc_hash::FxBuildHasher>,
+    inner: Option<QuickCache<CacheKey, Item, UnitWeighter, rustc_hash::FxBuildHasher>>,
```
marvin-j97: I would like the descriptor table to stay being as it was before, because there is nothing it really needs to do differently (single responsiblity).

src/compaction/leveled/mod.rs
```diff
@@ -20,7 +20,7 @@ fn pick_minimal_compaction(
     curr_run: &Run<Table>,
     next_run: Option<&Run<Table>>,
     hidden_set: &HiddenSet,
-    overshoot: u64,
+    _overshoot: u64,
```
marvin-j97: It would be nice to not have these clippy changes in this PR, or move them into another PR.

src/table/iter.rs
```diff
@@ -93,6 +93,7 @@ fn create_data_block_reader(block: DataBlock) -> OwnedDataBlockIter {
 pub struct Iter {
     table_id: GlobalTableId,
     path: Arc<PathBuf>,
+    pinned_file_descriptor: Option<Arc<File>>,
```
marvin-j97: Similar to the `load_block` thing, we really only have two cases:

1. we have a pinned file descriptor, or
2. a descriptor table

That could be modelled using an enum as well.

marvin-j97 (changes requested):

src/table/util.rs
kunalmohan: This definitely makes sense. Thanks for pointing it out! Looking through the code, IIUC, I can see that the `Table` struct only ever accesses its own FD from the DescriptorTable. Maybe we can store FileAccessor in there as well and abstract out DescriptorTable entirely (have access and insert definitions on the enum as well; insert will only act on the descriptor table variant and will be No-op for file)? What are your thoughts on this?

marvin-j97: That probably makes sense, so we don't have to actually handle both cases manually (so that no match is needed in `load_block`).

src/vlog/blob_file/writer.rs
```diff
@@ -12,6 +12,7 @@ use std::{
     fs::File,
     io::{BufWriter, Write},
     path::{Path, PathBuf},
+    sync::Arc,
```
marvin-j97: Import probably snuck in here by auto completion

PR #239, opened 2026-01-22
:recycle: Extract Filesystem IO trait

hhalex:


src/range.rs
```diff
@@ -65,39 +65,71 @@ pub fn prefix_to_range(prefix: &[u8]) -> (Bound<UserKey>, Bound<UserKey>) {
...
-    pub(crate) version: SuperVersion,
+pub struct IterState<F: FileSystem = crate::fs::StdFileSystem> {
+    pub(crate) version: SuperVersion<F>,
     pub(crate) ephemeral: Option<(Arc<Memtable>, SeqNo)>,
 }
 
 type BoxedMerge<'a> = Box<dyn DoubleEndedIterator<Item = crate::Result<InternalValue>> + Send + 'a>;
+type TreeIterJoinedCell<'a, F> =
+    self_cell::unsafe_self_cell::JoinedCell<IterState<F>, BoxedMerge<'a>>;
+
```
hhalex, line 76: Generics added forced me to expand the macro and customize

src/fs.rs
```diff
@@ -0,0 +1,120 @@
...
+    ffi::{OsStr, OsString},
+    fs,
+    io,
+    path::{Path, PathBuf},
+};
+
+/// Filesystem abstraction for pluggable storage backends.
+pub trait FileSystem: Send + Sync + std::panic::RefUnwindSafe + std::panic::UnwindSafe {
+    /// Opens an existing file for reading.
+    fn open(path: &Path) -> io::Result<fs::File>;
```
marvin-j97: While this trait allows creating a shim file system it still requires "proper" std::fs Files. For like really deterministic simulation testing and what not, File should really be a trait itself. Then the file system and actual files could be fully simulated, kind of like the backend trait in redb, though we have to define more stuff because we have to actually deal with directories.

hhalex: Yes totally agree, i am giving a try with another trait for Files. and add it as a type member of the FileSystem trait.

src/blob_tree/mod.rs
```diff
@@ -130,24 +131,32 @@ fn resolve_value_handle(
...
     #[doc(hidden)]
-    pub index: crate::Tree,
+    pub index: crate::Tree<F>,
 
     blobs_folder: Arc<PathBuf>,
 }
 
-impl BlobTree {
-    pub(crate) fn open(config: Config) -> crate::Result<Self> {
+impl<F: FileSystem> Clone for BlobTree<F> {
```
hhalex, line 138: deriving Clone apparently creates a constraint on F that must be Clonable

hhalex: I can put a Clone constraint on F, it seems hacky though

PR #241, opened 2026-02-04
Lower MSRV to 1.90

marceline-cramer:
See https://github.com/fjall-rs/fjall/issues/250

### DIFF
--- a/src/version/blob_file_list.rs
+++ b/src/version/blob_file_list.rs
@@ -1,8 +1,8 @@
 use crate::{
     blob_tree::FragmentationMap,
     vlog::{BlobFile, BlobFileId},
+    HashMap,
 };
-use std::collections::HashMap;
 
 #[derive(Clone, Default)]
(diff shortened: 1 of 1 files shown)


marvin-j97:
I think it would make more sense to see if the BTreeMap can be replaced with a HashMap. I don't actually remember what the rationale behind the B-tree was, but `HashMap::extract_if` is actually 1.88+. Though IIRC there was something else that required 1.89+.

marceline-cramer:
alright, i've replaced it with a HashMap and restored the `extract_if` syntax

src/version/blob_file_list.rs
```diff
@@ -2,13 +2,13 @@ use crate::{
     blob_tree::FragmentationMap,
     vlog::{BlobFile, BlobFileId},
 };
-use std::collections::BTreeMap;
+use std::collections::HashMap;
```
marvin-j97:
```suggestion
use crate::HashMap;
```

marceline-cramer: fixed

PR #242, opened 2026-02-06
feat: add range tombstone (delete_range) support

### LINKED ISSUE #2: Range tombstones/delete_range/delete_prefix
Use prefix .next and .next_back to get a  range for delete_prefix

temporaryfix:
## Summary

Implements native range tombstone support for the LSM-tree, enabling efficient deletion of contiguous key ranges without writing individual tombstones per key.

Closes #2

## Motivation

Currently, deleting a range of keys requires iterating over the range and writing a point tombstone for each key. This is expensive for large ranges — both in write amplification and in the tombstones that must be compacted away later. Range tombstones (as described in the [RocksDB DeleteRange design](https://github.com/facebook/rocksdb/wiki/DeleteRange-Implementation)) solve this by recording a single `[start, end)` interval with a sequence number, suppressing all keys within the range that have a lower seqno.

### Core types
- **`RangeTombstone`**: half-open `[start, end)` interval with seqno. Supports `contains_key`, `visible_at`, `should_suppress`, `intersect_opt`, and `fully_covers` queries. Ordered by `(start asc, seqno desc, end asc)`.
- **`ActiveTombstoneSet` / `ActiveTombstoneSetReverse`**: streaming sweep-line trackers for forward and reverse iteration, using a seqno multiset and min/max-heap expiry to efficiently determine suppression without rescanning all tombstones.
- **`CoveringRt`**: returned by table-skip queries to identify tombstones that fully cover a table's key range.

### Memtable integration
- **`IntervalTree`**: AVL-balanced BST keyed by start, augmented with `subtree_max_end` / `subtree_max_seqno` / `subtree_min_seqno` for efficient pruning. Supports `query_suppression`, `overlapping_tombstones`, and `covering_rt` queries.
- **Dual-indexed storage**: `IntervalTree` for point/overlap queries + `BTreeMap` by end-desc for reverse iteration.
- Methods: `insert_range_tombstone`, `is_suppressed_by_range_tombstone`, `overlapping_tombstones`, `range_tombstones_by_start/end_desc`.

### SST block format
- **Encoder**: `encode_by_start` (prefix-compressed START keys with per-window `max_end`) and `encode_by_end_desc` (prefix-compressed END keys). Both produce backward-parseable footer format with restart points.
- **ByStart decoder**: `query_suppression`, `overlapping_tombstones`, `query_covering_rt_for_range` with per-window `max_end` pruning for early exit.
- **ByEndDesc decoder**: `iter()` in end-descending order for reverse iteration.
- Two new `BlockType` variants: `RangeTombstoneStart` and `RangeTombstoneEnd`.

### Read path
- **Point reads**: after resolving a key from memtables or SSTs, check range tombstone suppression across all layers (active memtable, sealed memtables, SST tables). A key is suppressed if any visible tombstone covers it with a higher seqno.
- **Range/prefix iteration**: `RangeTombstoneFilter` wraps `MvccStream`, collecting tombstones from all sources and using `ActiveTombstoneSet` (forward) or `ActiveTombstoneSetReverse` (reverse) to suppress entries during iteration.
- **Table skipping**: during iteration setup, tables fully covered by a range tombstone (where tombstone seqno > table's max seqno) are skipped entirely, avoiding unnecessary I/O.

### Write path
- **Flush**: range tombstones from sealed memtables are collected and written into output SSTs via `MultiWriter`.
- **Compaction**: range tombstones from input tables are collected and written into output tables. Tombstones are clipped to each output table's key range via `intersect_opt()`. At the last compaction level, tombstones below the MVCC GC watermark are evicted since no data beneath them can be resurrected.

**Note on GC behavior:** After a range tombstone is evicted at the bottom level (when its seqno is below the GC watermark), the underlying data becomes visible again to new reads. Physical cleanup of suppressed keys happens lazily during future compactions when no snapshots hold them.

### MVCC correctness
- Tombstones respect snapshot visibility: `visible_at(read_seqno)` ensures only tombstones with `seqno < read_seqno` are considered.
- `should_suppress(key_seqno, read_seqno)` ensures a tombstone only suppresses keys with lower seqno than the tombstone itself.
- Tombstone eviction during compaction respects the GC watermark to preserve snapshot isolation.

## Test plan

- [x] 13 end-to-end integration tests (`tests/range_tombstone.rs`) covering:
  - Memtable point reads and range iteration
  - Flush persistence and recovery
  - Compaction propagation and tombstone clipping
  - End-exclusive `[start, end)` semantics
  - Reverse iteration
  - MVCC snapshot visibility
  - Overlapping tombstones
  - Cross-layer suppression (memtable tombstone suppressing SST data)
  - GC threshold eviction (data becomes visible after tombstone eviction)
  - Table skip optimization during iteration
- [x] Unit tests for `IntervalTree`, `ActiveTombstoneSet`, `RangeTombstone`, block encoders/decoders
- [x] Full `cargo test` passes (312 unit + 23 doc tests)

## Stats

- **23 files changed**, ~3,900 lines added
- 12 commits, incrementally buildable

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tests/range_tombstone.rs
+++ b/tests/range_tombstone.rs
@@ -1,4 +1,4 @@
-use lsm_tree::{AbstractTree, Config, Guard, RangeTombstone, SeqNo, SequenceNumberCounter};
+use lsm_tree::{AbstractTree, Config, Guard, SeqNo, SequenceNumberCounter};
 
 fn open_tree(folder: &std::path::Path) -> lsm_tree::AnyTree {
     Config::new(
@@ -22,8 +22,7 @@ fn range_tombstone_memtable_point_read() -> lsm_tree::Result<()> {
     }
 
     // Insert range tombstone [b, e) at seqno 10
-    tree.active_memtable()
-        .insert_range_tombstone(RangeTombstone::new("b".into(), "e".into(), 10));
+    tree.remove_range("b", "e", 10);
 
     // "a" should be visible (outside range)
     assert!(tree.get("a", SeqNo::MAX)?.is_some());
@@ -47,8 +46,7 @@ fn range_tombstone_memtable_range_iter() -> lsm_tree::Result<()> {
         tree.insert(*key, "val", i as SeqNo);
     }
 
-    tree.active_memtable()
-        .insert_range_tombstone(RangeTombstone::new("b".into(), "e".into(), 10));
+    tree.remove_range("b", "e", 10);
 
     let keys: Vec<_> = tree
         .range::<&str, _>(.., SeqNo::MAX, None)
@@ -73,8 +71,7 @@ fn range_tombstone_survives_flush() -> lsm_tree::Result<()> {
         tree.insert(*key, "val", i as SeqNo);
     }
 
-    tree.active_memtable()
-        .insert_range_tombstone(RangeTombstone::new("b".into(), "e".into(), 10));
+    tree.remove_range("b", "e", 10);
 
     tree.flush_active_memtable(0)?;
     assert_eq!(1, tree.table_count());
@@ -103,8 +100,7 @@ fn range_tombstone_survives_compaction() -> lsm_tree::Result<()> {
         tree.insert(*key, "val", i as SeqNo);
     }
 
-    tree.active_memtable()
-        .insert_range_tombstone(RangeTombstone::new("b".into(), "e".into(), 10));
+    tree.remove_range("b", "e", 10);
 
     tree.flush_active_memtable(0)?;
     tree.major_compact(u64::MAX, 0)?;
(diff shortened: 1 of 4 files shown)


tests/range_tombstone.rs
```diff
@@ -340,3 +340,42 @@ fn range_tombstone_table_skip_in_iteration() -> lsm_tree::Result<()> {
...
+    let folder = tempfile::tempdir()?;
+    let tree = open_tree(folder.path());
+
+    // Insert data at low seqnos
+    tree.insert("a", "val_a", 1);
+    tree.insert("b", "val_b", 2);
+    tree.insert("c", "val_c", 3);
+
+    // Insert range tombstone [a, d) at seqno 10
+    tree.active_memtable()
```
marvin-j97: AbstractTree should then have a remove_range method, so we don't have to access the active memtable manually.

PR #287, opened 2026-04-18
Provide an `as_slice` method on the Slice type

irevoire:
Hey!

I made this small helper similar to [the one in the stdlib](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.as_slice) after having to write `buf.extend(&slice as &[u8]);` a bunch of times.

## Summary by CodeRabbit

* **New Features**
  * Added a public read-only accessor that returns the full underlying byte data as a borrowed slice, enabling direct, zero-copy reads.
* **Tests**
  * Added a unit test confirming the accessor returns the original byte data unchanged.

### DIFF
--- a/src/slice/slice_default/mod.rs
+++ b/src/slice/slice_default/mod.rs
@@ -22,7 +22,7 @@ impl Slice {
     /// Returns the bytes slice containing the entire data.
     #[must_use]
     pub fn as_slice(&self) -> &[u8] {
-        &self
+        self
     }
 
     #[doc(hidden)]
--- a/src/slice/slice_default/mod.rs
+++ b/src/slice/slice_default/mod.rs
@@ -19,6 +19,12 @@ impl Slice {
         Self(bytes.into())
     }
 
+    /// Returns the bytes slice containing the entire data.
+    #[must_use]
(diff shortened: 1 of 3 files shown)


marvin-j97:
This should probably be implemented for Slice in general, not just the `byteview` flavour.

src/slice/slice_default/mod.rs
```diff
@@ -19,6 +19,12 @@ impl Slice {
         Self(bytes.into())
     }
 
+    /// Returns the bytes slice containing the entire data.
+    #[must_use]
+    pub fn as_slice(&self) -> &[u8] {
+        &self
+    }
+
```
marvin-j97, line 27: This is currently missing for the Bytes-backed Slice implementation.

marvin-j97 (changes requested):

irevoire:
Thanks for the review!
I'm not super available right now but will look into it in a few days!

src/slice/slice_default/mod.rs
irevoire: Done, since the change was minor I did a squash

marvin-j97 (approved):

PR #288, opened 2026-04-19
Remove/update benchmarks

kunalmohan:
The benchmarks were outdated and did not compile. This PR aims to update or remove the benchmarks.

## Summary by CodeRabbit

* **Chores**
  * Removed several legacy benchmark targets and retired multiple internal performance benchmarks to reduce maintenance.
  * Simplified benchmark registrations and groups.

* **Tests**
  * Updated remaining benchmark suites to use current initialization and API patterns for consistency (explicit constructors, unified cache/sequence handling).
  * Removed redundant or deprecated benchmark cases.

PR #290, opened 2026-04-26
Feat: Add batch write implementation

### LINKED ISSUE #283: Add write_batch method to AbstractTree
Currently in fjall, batches have to loop over the WriteBatch and conditionally call` tree.insert`/`remove`/`remove_weak`, which delegate to `append_entry`. `append_entry` costs at least one atomic op, in reality though more (because of `super_version.latest_version`). For large batch inserts and transactions this is unnecessary work that can be skipped by introducing a new method that gets the current version once and then inserts an iterator of items into the active memtable instead.

## Possible API

`AbstractTree::write_batch(it: impl Iterator<Item = BatchItem>) -> (u64, u64)`

Return type should be same as a single append_entry, but the size sum of all inserted entries plus the new memtable size.

## What to do

1. Implement `AbstractTree::write_batch`
2. Benchmark write batches of different sizes with new implementation against the naive `for-loop + append_entry` implementation... (is it worth it?)

kunalmohan:
Fix #283 

Note: This builds on #288 

This is a two step implementation:
- A batch_write in `AbstractTree` which calls `memtable::insert()` in a loop.
- A further optimization ([ad76efc266d562a97903d8e7c853701c8bb89d22](https://github.com/fjall-rs/lsm-tree/commit/ad76efc266d562a97903d8e7c853701c8bb89d22)) attempt is made by adding a new `memtable::insert_batch()` which loops on the items internally but performs atomic replace only once in the end.

Results from benchmarks:

Results do not look promising: There is very minimal performace improvement with both the optimizations.

memtable::insert() loop:
Screenshot 2026-04-26 at 11 21 29 PM

memtable::insert_batch():
Screenshot 2026-04-26 at 11 19 54 PM

## Summary by CodeRabbit

* **New Features**
  * Added batch write support to tree operations with a new `BatchItem` type, enabling multiple insert and delete operations to be submitted in a single batch call, improving write efficiency.


kunalmohan:
Updated benchmark:
(Still not much improvement)

Screenshot 2026-04-26 at 11 53 02 PM

PR #293, opened 2026-06-01
use the SuperVersion seqno for get_highest_persisted_seqno if the tree is totally empty

### LINKED ISSUE #288: idle keyspaces leads to unbounded sealed journals
**Describe the bug**
We've noticed that our application tends to accumulate journal segments; every once in a while, a sealed journal will get "stuck" and won't get cleared after recovery. Once this happens, of course, no subsequent sealed journals can ever get cleared so the journal size grows without bound until the application gets so bloated that startup can't finish within the kubernetes startup probe timeout and it breaks.

In the cases I've examined, the journal contains a record pointing to a keyspace that isn't actively being used on this instance, whose memtable is empty. In this case, the table can never be flushed (the `super_version.active_memtable.is_empty()` call in lsm-tree return false), so we can never actually increase the persisted seqno and never get the journal out. One possible way to create this scenario is to make a call to `clear()` for a keyspace, but do no subsequent writes. It's still tricky to do this because you need to end up sealing that journal (so you need to outrun the flush thread).

<details>

<summary>some example code to recreate this</summary>

The following code spams lots of writes combined with interspersed `.clear()` calls to recreate this:

```rust
use anyhow::Context;
use std::sync::{
    Arc,
    atomic::{AtomicU32, Ordering},
};

fn main() -> anyhow::Result<()> {
    env_logger::init();

    let db = fjall::Database::builder("./db")
        // we need to get past the 64MB threshold to ensure that the journal gets sealed
        .max_journaling_size(72 * 1024 * 1024)
        // limit the size of the worker pool
        .worker_threads(1)
        // disable automatic DB persistence to prevent a barrier between the clear and the writes
        .manual_journal_persist(true)
        .open()
        .context("opening database")?;

    let cold_keyspace = db.keyspace("cold", || {
        fjall::KeyspaceCreateOptions::default().manual_journal_persist(true)
    })?;

    // run a bunch of threads not because of concurrency, but because we need to try and
    // outrace the worker pool
    std::thread::scope(|spawner| {
        let done = Arc::new(AtomicU32::new(0));

        spawner.spawn({
            let cold_keyspace = cold_keyspace.clone();
            let done = done.clone();
            move || -> anyhow::Result<()> {
                while done.load(Ordering::Relaxed) < 10 {
                    cold_keyspace.clear()?;
                    std::thread::sleep(std::time::Duration::from_millis(500));
                }
                Ok(())
            }
        });

        for k_i in 0..10 {
            let db = db.clone();
            let name = format!("hot{k_i}");
            log::debug!("spawning writer for {name}");
            let keyspace = db
                .keyspace(&name, fjall::KeyspaceCreateOptions::default)
                .unwrap();
            let done = done.clone();
            spawner.spawn(move || -> anyhow::Result<()> {
                for i in 0..=(128 * 1024) {
                    let mut batch = db.batch().durability(Some(fjall::PersistMode::Buffer));
                    for j in 0..128 {
                        let idx: u64 = i * 128 + j;
                        let key = idx.to_be_bytes();
                        batch.insert(&keyspace, key, [(idx % 256) as u8; 64]);
                    }
                    batch.commit()?;
                }
                done.fetch_add(1, Ordering::Relaxed);
                Ok(())
            });
        }
    });

    Ok(())
}
```

</details>

I'm not immediately sure what the best fix for this is. Maybe when replaying sealed journals, fjall should force the lsm tree to do a rotation with an empty tree? Or maybe lsm-tree should consider the sequence number of the most recent version itself in `get_highest_persisted_seqno()`? I have an `lsm-tree` patch that makes `clear()` take a SeqNo argument and just stores it in memory and then regurgitates it in `get_last_persisted_seqno`, which does resolve this issue but I'm not sure if it's the best way to do so.

svix-jbrown:
I talked to @marvin-j97  in discord and they suggest that this would be a simpler/better alternative to #291.

If a tree is totally empty (no tables on disk, nothing in the memtables; e.g., the state right after a clear), then we use the seqno of the SuperVersion as the highest persisted seqno.

I believe that this will fix fjall-rs/fjall#288

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable tracking of a tree's highest persisted sequence number so open, insert, clear, and reopen report consistent persisted-seqno state.

* **New Features**
  * Trees now record and restore an optional creation seqno and initialize from the configured start seqno.
  * Drop/compaction now records explicit seqno markers for persisted history.

* **Tests**
  * Strengthened tests validating persisted-seqno transitions across insert, clear, persist, and reopen.

### DIFF
--- a/src/compaction/worker.rs
+++ b/src/compaction/worker.rs
@@ -612,9 +612,11 @@ fn drop_tables(
         |current| {
             let mut copy = current.clone();
 
-            copy.version = copy
-                .version
-                .with_dropped(ids_to_drop, &mut dropped_blob_files)?;
+            copy.version = copy.version.with_dropped_at(
+                ids_to_drop,
+                &mut dropped_blob_files,
+                Some(opts.global_seqno.get()),
+            )?;
 
             Ok(copy)
         },
--- a/src/tree/inner.rs
+++ b/src/tree/inner.rs
@@ -73,13 +73,14 @@ pub struct TreeInner {
 
 impl TreeInner {
     pub(crate) fn create_new(config: Config) -> crate::Result<Self> {
-        let version = Version::new(
+        let version = Version::new_at_seqno(
             0,
             if config.kv_separation_opts.is_some() {
                 crate::TreeType::Blob
             } else {
                 crate::TreeType::Standard
             },
+            Some(config.seqno.get()),
         );
         persist_version(&config.path, &version)?;
 
--- a/src/tree/mod.rs
+++ b/src/tree/mod.rs
@@ -271,7 +271,11 @@ impl AbstractTree for Tree {
                 let mut copy = v.clone();
                 copy.active_memtable = Arc::new(Memtable::new(self.memtable_id_counter.next()));
                 copy.sealed_memtables = Arc::default();
-                copy.version = Version::new(v.version.id() + 1, self.tree_type());
+                copy.version = Version::new_at_seqno(
+                    v.version.id() + 1,
+                    self.tree_type(),
+                    Some(config.seqno.get()),
+                );
                 Ok(copy)
             },
             &config.seqno,
@@ -630,10 +634,17 @@ impl AbstractTree for Tree {
     }
 
     fn get_highest_persisted_seqno(&self) -> Option<SeqNo> {
-        self.current_version()
-            .iter_tables()
-            .map(Table::get_highest_seqno)
-            .max()
+        let guard = self.version_history.read().expect("lock is poisoned");
+        let super_version = guard.latest_version();
+        if super_version.is_empty() {
+            super_version.version.created_seqno
+        } else {
+            super_version
+                .version
+                .iter_tables()
+                .map(Table::get_highest_seqno)
+                .max()
+        }
     }
 
     fn get<K: AsRef<[u8]>>(&self, key: K, seqno: SeqNo) -> crate::Result<Option<UserValue>> {
--- a/src/tree/mod.rs
+++ b/src/tree/mod.rs
@@ -634,8 +634,12 @@ impl AbstractTree for Tree {
     }
 
     fn get_highest_persisted_seqno(&self) -> Option<SeqNo> {
-        let guard = self.version_history.read().expect("lock is poisoned");
-        let super_version = guard.latest_version();
+        #[expect(clippy::expect_used, reason = "lock is expected to not be poisoned")]
+        let super_version = self
+            .version_history
+            .read()
+            .expect("lock is poisoned")
+            .latest_version();
         if super_version.is_empty() {
             super_version.version.created_seqno
         } else {
--- a/src/version/mod.rs
+++ b/src/version/mod.rs
@@ -19,11 +19,11 @@ use crate::checksum::ChecksumType;
 use crate::coding::Encode;
 use crate::compaction::state::hidden_set::HiddenSet;
 use crate::version::recovery::Recovery;
-use crate::TreeType;
 use crate::{
     vlog::{BlobFile, BlobFileId},
     HashSet, KeyRange, Table, TableId,
 };
+use crate::{SeqNo, TreeType};
 use optimize::optimize_runs;
 use run::Ranged;
 use std::{ops::Deref, sync::Arc};
@@ -158,6 +158,9 @@ pub struct VersionInner {
 
     /// Blob file fragmentation
     gc_stats: Arc<FragmentationMap>,
+
+    /// seqno at which this version was born
+    pub created_seqno: Option<SeqNo>,
 }
 
 /// A version is an immutable, point-in-time view of a tree's structure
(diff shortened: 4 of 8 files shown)


src/tree/mod.rs
```diff
@@ -630,10 +634,17 @@ impl AbstractTree for Tree {
     }
 
     fn get_highest_persisted_seqno(&self) -> Option<SeqNo> {
-        self.current_version()
-            .iter_tables()
-            .map(Table::get_highest_seqno)
-            .max()
+        let guard = self.version_history.read().expect("lock is poisoned");
+        let super_version = guard.latest_version();
```
marvin-j97:
```suggestion
        let super_version = self.current_version();
```

src/version/mod.rs
```diff
@@ -158,6 +158,9 @@ pub struct VersionInner {
 
     /// Blob file fragmentation
     gc_stats: Arc<FragmentationMap>,
+
+    /// seqno at which this version was born
```
marvin-j97:
```suggestion
    /// Seqno at which this version was created
    ///
    /// Note: This is optional because older lsm-tree versions did not persist the version seqno.
```

src/tree/mod.rs
svix-jbrown: doesn't quite work because `current_version()` returns the `Version`, but information about the memtables is only on the `SuperVersion`.

src/version/mod.rs
```diff
@@ -468,13 +496,22 @@ impl Version {
             Arc::new(copy)
         };
 
+        // reset the created_seqno if the tree has newly become empty
+        let table_count: usize = levels.iter().map(|x| x.table_count()).sum();
+        let created_seqno = if table_count == 0 {
+            self.created_seqno.max(seqno)
+        } else {
+            self.created_seqno
+        };
```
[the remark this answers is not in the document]
svix-jbrown: sequence numbers can't go backwards, silly robot

src/version/mod.rs
```diff
@@ -209,6 +215,11 @@ impl Version {
 
     /// Creates a new empty version.
     pub fn new(id: VersionId, tree_type: TreeType) -> Self {
+        Self::new_at_seqno(id, tree_type, None)
+    }
```
marvin-j97, line 219: This may not be needed anymore - we always build Versions with seqno now anyway?

src/version/mod.rs
```diff
@@ -209,6 +215,11 @@ impl Version {
...
+
+    /// Creates a new empty version at the given sequence number
+    pub fn new_at_seqno(id: VersionId, tree_type: TreeType, created_seqno: Option<SeqNo>) -> Self {
```
marvin-j97, line 222: `created_seqno: Option<SeqNo>` -> `created_seqno: SeqNo`

src/compaction/worker.rs
```diff
@@ -612,9 +612,11 @@ fn drop_tables(
         |current| {
             let mut copy = current.clone();
 
-            copy.version = copy
-                .version
-                .with_dropped(ids_to_drop, &mut dropped_blob_files)?;
+            copy.version = copy.version.with_dropped_at(
+                ids_to_drop,
+                &mut dropped_blob_files,
+                Some(opts.global_seqno.get()),
```
marvin-j97, line 618: ~~I think this should be `opts.global_seqno.next()`. Otherwise two compactions being installed in quick succession could, technically, have the same seqno.~~

Ignore

src/tree/inner.rs
```diff
@@ -73,13 +73,14 @@ pub struct TreeInner {
...
     pub(crate) fn create_new(config: Config) -> crate::Result<Self> {
-        let version = Version::new(
+        let version = Version::new_at_seqno(
             0,
             if config.kv_separation_opts.is_some() {
                 crate::TreeType::Blob
             } else {
                 crate::TreeType::Standard
             },
+            Some(config.seqno.get()),
```
marvin-j97, line 83: ~~Same here.~~

src/tree/mod.rs
```diff
@@ -271,7 +271,11 @@ impl AbstractTree for Tree {
                 let mut copy = v.clone();
                 copy.active_memtable = Arc::new(Memtable::new(self.memtable_id_counter.next()));
                 copy.sealed_memtables = Arc::default();
-                copy.version = Version::new(v.version.id() + 1, self.tree_type());
+                copy.version = Version::new_at_seqno(
+                    v.version.id() + 1,
+                    self.tree_type(),
+                    Some(config.seqno.get()),
```
marvin-j97, line 277: ~~And here.~~

src/version/mod.rs
```diff
@@ -468,13 +496,22 @@ impl Version {
             Arc::new(copy)
         };
 
+        // reset the created_seqno if the tree has newly become empty
+        let table_count: usize = levels.iter().map(|x| x.table_count()).sum();
+        let created_seqno = if table_count == 0 {
+            self.created_seqno.max(seqno)
```
marvin-j97, line 502: Is this even possible? Why should the persisted seqno be possibly higher than the next, newer version?

src/version/mod.rs
svix-jbrown: I wasn't sure if you wanted to change the public API for this library, since it might have users outside of fjall; hence adding all the new methods

marvin-j97: Version is not exposed as public either way

PR #301, opened 2026-07-05
fix: persist correct compression type for relocated blob files

### LINKED ISSUE #300: Blob files written by relocating compaction persist wrong compression type, corrupting reads after reopen
**Version:** lsm-tree 3.1.5 (via fjall 3.1.5), blob tree with `KvSeparationOptions::default()` (LZ4)

## Symptom

After days of normal operation (with restarts in between), a point read on a blob tree panics:

```
thread '<unnamed>' panicked at lsm-tree-3.1.5\src\vlog\blob_file\reader.rs:98:9:
assertion `left == right` failed
  left: 1437
 right: 1015
```

`left` is `real_val_len` from the blob header (uncompressed length), `right` is the value returned by the reader — which equals the vhandle's `on_disk_size`, i.e. the LZ4-compressed length. The blob's checksum verifies fine. In release builds the `debug_assert_eq!` compiles out and the **LZ4-compressed bytes are silently returned to the application as the value**.

## Cause

When a compaction relocates blobs out of fragmented blob files (`RelocatingCompaction`), the blob writer is set up for passthrough — compressed blobs are copied byte-for-byte, and the compression type is only stamped on metadata:

```rust
// compaction/worker.rs
let writer = BlobFileWriter::new(..)?
    .use_target_size(blob_opts.file_target_size)
    .use_passthrough_compression(blob_opts.compression);
```

However, `use_passthrough_compression` only affects the **in-memory** `BlobFile` handed out by `MultiWriter::consume_writer`:

```rust
// vlog/blob_file/multi_writer.rs
compression: if passthrough_compression == CompressionType::None {
    metadata.compression
} else {
    passthrough_compression // <- patched in memory only
},
```

The metadata section **persisted to disk** is written earlier by `Writer::finish()`, which uses the inner writer's own compression — `CompressionType::None` for the passthrough writer:

```rust
// vlog/blob_file/writer.rs (finish)
let metadata = Metadata {
    ..
    compression: self.compression, // None — passthrough never sets this
};
metadata.encode_into(&mut self.writer)?;
```

So the file contains LZ4-compressed blobs but its persisted metadata says `None`. Reads keep working until the keyspace is reopened; recovery (`recover_blob_files` → `Metadata::from_slice`) then trusts the persisted `None`, and `Reader::get` takes the no-decompression branch, returning raw LZ4 bytes (checksum passes, since it covers the on-disk bytes).

## Reproduction sketch

1. Open a blob tree with LZ4 blob compression; write enough separated values to create blob files.
2. Overwrite/delete enough of them so a compaction triggers blob file relocation (`RelocatingCompaction`).
3. Reopen the tree.
4. `get` a relocated key → debug: panic at `vlog/blob_file/reader.rs:98`; release: returns compressed garbage.

## Suggested fix

Persist the effective compression type: when finishing a passthrough writer, encode the passthrough compression into the on-disk metadata (e.g. propagate `passthrough_compression` into `Writer` / `Writer::finish()`) instead of the writer's own `compression`. Existing blob files written by relocation are already on disk with the wrong metadata, so recovery may also want a fallback (blob headers carry both lengths, so a mismatch is detectable).

styris-ame:
Blob files written during blob relocation stored CompressionType::None in their meta block, so compressed blobs were returned without decompression after recovery.

Fixes https://github.com/fjall-rs/lsm-tree/issues/300

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved blob compression handling so stored data now preserves the expected compression settings through rotation and file finalization.
  * Fixed recovery behavior around major compaction, helping ensure values remain accessible after blob relocation and reopen.

### DIFF
--- a/src/vlog/blob_file/multi_writer.rs
+++ b/src/vlog/blob_file/multi_writer.rs
@@ -84,6 +84,7 @@ impl MultiWriter {
     pub(crate) fn use_passthrough_compression(mut self, compression: CompressionType) -> Self {
         assert_eq!(self.compression, CompressionType::None);
         self.passthrough_compression = compression;
+        self.active_writer.passthrough_compression = compression;
         self
     }
 
@@ -104,14 +105,11 @@ impl MultiWriter {
         let blob_file_path = self.folder.join(new_blob_file_id.to_string());
 
         let new_writer = Writer::new(blob_file_path, new_blob_file_id, self.tree_id)?
-            .use_compression(self.compression);
+            .use_compression(self.compression)
+            .use_passthrough_compression(self.passthrough_compression);
 
         let old_writer = std::mem::replace(&mut self.active_writer, new_writer);
-        let blob_file = Self::consume_writer(
-            old_writer,
-            self.passthrough_compression,
-            self.descriptor_table.clone(),
-        )?;
+        let blob_file = Self::consume_writer(old_writer, self.descriptor_table.clone())?;
         self.results.extend(blob_file);
 
         Ok(())
@@ -119,7 +117,6 @@ impl MultiWriter {
 
     fn consume_writer(
         writer: Writer,
-        passthrough_compression: CompressionType,
         descriptor_table: Option<Arc<DescriptorTable>>,
     ) -> crate::Result<Option<BlobFile>> {
         if writer.item_count > 0 {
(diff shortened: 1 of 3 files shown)


marvin-j97:
Thanks!
(That code may need some refactor in the future, having two compression variables should probably be an enum)

PR #305, opened 2026-07-11
fix: retry transient file errors on Windows

### LINKED ISSUE #304: Transient Windows file errors crash compaction workers and poison the database
**Version:** lsm-tree 3.1.6 / fjall 3.1.6, Windows

## What happened

During a heavy write load on Windows, a compaction worker hit a transient `Access is denied` while committing a version change. On Windows, renames and deletes fail with a sharing violation if any other program (antivirus, indexer, backup agent) briefly holds the file open without share-delete. Since the error escapes the worker, fjall poisons the whole database over a hiccup that would have succeeded milliseconds later.

```
ERROR fjall::compaction::worker: Compaction failed: Io(Os { code: 5, kind: PermissionDenied, message: "Access is denied." })
ERROR fjall::worker_pool: Worker #3 crashed: Storage(Io(Os { code: 5, kind: PermissionDenied, message: "Access is denied." }))
ERROR fjall::compaction::worker: Compaction failed: Io(Os { code: 80, kind: AlreadyExists, message: "The file exists." })
ERROR fjall::worker_pool: Worker #1 crashed: Storage(Io(Os { code: 80, kind: AlreadyExists, message: "The file exists." }))
failed to persist 25 log entries: FjallError: Poisoned
```

There are two separate problems here:

1. **No tolerance for transient sharing violations.** The rename in `rewrite_atomic` (`NamedTempFile::persist`, i.e. `MoveFileExW`) and the manifest delete in `SuperVersions::maintenance` bubble a one-off `ERROR_ACCESS_DENIED` straight up as a worker crash. There is even a `// TODO: not sure why it fails on Windows...` in `rewrite_atomic` pointing at this. POSIX is unaffected since rename/unlink don't care about open handles.

2. **A failed persist leaves a partial version file that wedges all later ones.** `persist_version` does `File::create_new("v{N+1}")` and then writes into it. If anything after the create fails, the partial `v{N+1}` stays on disk while the in-memory version stays at `N`, so every following version upgrade recomputes `N+1` and dies with `AlreadyExists`. That is the worker #1/#2 cascade above; it is cross-platform (e.g. reachable via ENOSPC on Linux). Recovery cleans orphans up at startup, so it self-heals on restart, but at runtime one transient error turns into a permanent crash loop until the process restarts.

## Fix

- Retry renames/deletes on Windows sharing violations (`ERROR_ACCESS_DENIED`, `ERROR_SHARING_VIOLATION`, `ERROR_LOCK_VIOLATION`, `ERROR_USER_MAPPED_FILE`) with bounded exponential backoff. This is the same approach RocksDB takes in its Windows port. No behavior change on other platforms.
- Write version manifests to a temp file and rename into place, so a failed persist leaves nothing at the final name and an orphan from a crash gets overwritten instead of colliding. Also flushes and syncs the file before the rename and fsyncs the directory after it.

styris-ame:
External tools briefly holding files open can fail renames and deletes in the compaction worker, poisoning the database. Retry with backoff, the same approach RocksDB uses on Windows. Version manifests are also written to a temp file and renamed into place now, so a failed persist leaves no partial file behind to collide with the next one.

Fixes https://github.com/fjall-rs/lsm-tree/issues/304

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows reliability for atomic file updates by retrying transient sharing/locking-related I/O failures with exponential backoff.
  * Strengthened version file persistence by using a safer temp-file finalization flow and ensuring data is fully flushed and synced before publishing.
  * Version files now correctly replace any existing or partial destination artifacts after persistence.
  * Maintenance cleanup now retries transient failures when removing outdated version files, improving robustness.

### DIFF
--- a/src/file.rs
+++ b/src/file.rs
@@ -58,6 +58,56 @@ pub fn read_exact(file: &File, offset: u64, size: usize) -> std::io::Result<Slic
     Ok(builder.freeze().into())
 }
 
+/// Runs an I/O operation, retrying transient Windows errors with backoff.
+///
+/// External programs (antivirus, indexers, backup agents) may briefly hold files open
+/// without sharing flags on Windows, failing renames and deletes that would succeed a
+/// moment later; POSIX has no such failure mode, so elsewhere the operation runs once.
+pub fn retry_transient_io<T>(mut op: impl FnMut() -> std::io::Result<T>) -> std::io::Result<T> {
+    #[cfg(windows)]
+    {
+        const MAX_ATTEMPTS: u32 = 10;
+
+        let mut delay = std::time::Duration::from_millis(1);
+
+        for _ in 1..MAX_ATTEMPTS {
+            match op() {
+                Err(e) if is_transient_windows_error(&e) => {
+                    std::thread::sleep(delay);
+                    delay *= 2;
+                }
+                result => return result,
+            }
+        }
+    }
+
+    op()
+}
+
+#[cfg(windows)]
+fn is_transient_windows_error(e: &std::io::Error) -> bool {
+    // ERROR_ACCESS_DENIED, ERROR_SHARING_VIOLATION, ERROR_LOCK_VIOLATION, ERROR_USER_MAPPED_FILE
+    matches!(e.raw_os_error(), Some(5 | 32 | 33 | 1224))
+}
+
+/// Persists a named temporary file to its final path, replacing any existing file.
+pub fn persist_temp_file(temp_file: tempfile::NamedTempFile, path: &Path) -> std::io::Result<()> {
+    let mut temp_file = Some(temp_file);
+
+    retry_transient_io(|| {
+        #[expect(clippy::expect_used, reason = "the temp file is put back on failure")]
+        temp_file
+            .take()
+            .expect("temp file should be present")
+            .persist(path)
+            .map(|_| ())
+            .map_err(|e| {
+                temp_file = Some(e.file);
+                e.error
+            })
+    })
+}
+
 /// Atomically rewrites a file.
 pub fn rewrite_atomic(path: &Path, content: &[u8]) -> std::io::Result<()> {
     #[expect(
@@ -70,7 +120,7 @@ pub fn rewrite_atomic(path: &Path, content: &[u8]) -> std::io::Result<()> {
     temp_file.write_all(content)?;
     temp_file.flush()?;
     temp_file.as_file_mut().sync_all()?;
-    temp_file.persist(path)?;
+    persist_temp_file(temp_file, path)?;
 
     // TODO: not sure why it fails on Windows...
     #[cfg(not(target_os = "windows"))]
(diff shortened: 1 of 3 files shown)


src/file.rs
```diff
@@ -58,6 +58,58 @@ pub fn read_exact(file: &File, offset: u64, size: usize) -> std::io::Result<Slic
...
+        match temp_file
+            .take()
+            .expect("temp file should be present")
+            .persist(path)
+        {
+            Ok(_) => Ok(()),
+            Err(e) => {
+                temp_file = Some(e.file);
+                Err(e.error)
+            }
```
marvin-j97: I think this can be refactored using `inspect_err`

styris-ame: `inspect_err` only gives a shared reference to the `PersistError`, but the code needs to move e.file out of it to put the temp file back for the next retry attempt.

However this could be done if you prefer it:
```rust
        temp_file
            .take()
            .expect("temp file should be present")
            .persist(path)
            .map(|_| ())
            .map_err(|e| {
                temp_file = Some(e.file);
                e.error
            })
```

Just let me know.

marvin-j97: I think the issue is more this taking out/putting into Option thing.
It can probably solved in a nicer way by just using `Option::as_mut?

Edit: Actually I don't get the Option at all

Suggestion:

```
pub fn persist_temp_file(mut temp_file: tempfile::NamedTempFile, path: &Path) -> std::io::Result<()> {
    retry_transient_io(|| {
        if let Err(e) = temp_file.persist(path) {
            temp_file = e.file;
            Err(e.error)
        } else {
            Ok(())
        }
    })
}
```

src/file.rs
```diff
@@ -58,6 +58,58 @@ pub fn read_exact(file: &File, offset: u64, size: usize) -> std::io::Result<Slic
...
+/// without sharing flags on Windows, failing renames and deletes that would succeed a
+/// moment later; POSIX has no such failure mode, so elsewhere the operation runs once.
+pub fn retry_transient_io<T>(mut op: impl FnMut() -> std::io::Result<T>) -> std::io::Result<T> {
+    #[cfg(windows)]
+    {
+        const MAX_ATTEMPTS: u32 = 10;
+
+        let mut delay = std::time::Duration::from_millis(1);
+
+        for _ in 1..MAX_ATTEMPTS {
```
marvin-j97, line 73: Should be 0..MAX_ATTEMPTS ... or 1..=MAX_ATTEMPTS

styris-ame: There's an unconditional final `op()` after the loop (it doubles as the sole attempt on non-Windows). So the total is 9 + 1 = 10 attempts = MAX_ATTEMPTS, with correctly no sleep after the last attempt. Changing it to 0..MAX_ATTEMPTS or 1..=MAX_ATTEMPTS would make 11 attempts.

src/file.rs
styris-ame: Doesn't compile. `NamedTempFile::persist` takes self by value, so calling it moves `temp_file` out of the closure capture, and moving out of a capture is not allowed in a `FnMut` closure, even though the `Err` arm puts it back. The Option is what makes the move legal, via `take()`.

marvin-j97 (approved):

src/version/persist.rs
```diff
@@ -13,9 +16,10 @@ pub fn persist_version(folder: &Path, version: &Version) -> crate::Result<()> {
...
     );
 
-    let path = folder.join(format!("v{}", version.id()));
-    let file = std::fs::File::create_new(path)?;
-    let writer = BufWriter::new(file);
+    // Written to a temporary file first: a failed persist then leaves no partial version
+    // file behind, which would otherwise collide with the next persist of the same ID.
+    let temp_file = tempfile::NamedTempFile::new_in(folder)?;
+    let writer = BufWriter::new(temp_file.as_file());
     let mut writer = ChecksummedWriter::new(writer);
```
marvin-j97, line 22: Instead of a tempfile, `create_new` could maybe just be switched to `create` because that truncates the existing file anyways.

styris-ame: Done

src/version/persist.rs
```diff
@@ -27,13 +30,19 @@ pub fn persist_version(folder: &Path, version: &Version) -> crate::Result<()> {
...
+    writer.flush()?;
+
     let checksum = writer.checksum();
 
+    drop(writer);
+
+    file.sync_all()?;
+
+    // IMPORTANT: fsync folder on Unix
+    fsync_directory(folder)?;
```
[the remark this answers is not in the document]
styris-ame: `fsync_directory` is already `cfg`-gated in `src/file.rs`, and is a no-op on Windows.

marvin-j97 (approved):

Issue #310, opened 2026-07-23
Triggering PoisonError in FIFO compaction when inserting data


StephanvanSchaik:
**Describe the bug**

I have an existing fjall database of about 89 GB where two of the partitions have the compaction strategy set to FIFO (with 16M as limit for both). I suspect these partitions are causing problems due to where the lsm-tree crate starts panicking.

```
thread 'fjall:worker' (12888) panicked at /home/stephan/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/lsm-tree-3.1.6/src/compaction/fifo.rs:82:9:
L0 needs to be disjoint
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

thread 'fjall:worker' (12887) panicked at /home/stephan/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/lsm-tree-3.1.6/src/compaction/worker.rs:94:57:
lock is poisoned: PoisonError { .. }

thread 'fjall:worker' (12886) panicked at /home/stephan/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/lsm-tree-3.1.6/src/compaction/worker.rs:94:57:
lock is poisoned: PoisonError { .. }
```

**To Reproduce**

So I am not fully sure how to reproduce this, but in my case inserting enough data into any partition, including non-FIFO partitions, seems to trigger the panic for me now, which I think just triggers the compaction process. The FIFO partitions also have data inserted into them over time.

**Expected behavior**

**Desktop:**
 - OS: Gentoo Linux (Linux 6.17.1)
 - Rust version: 1.96.1 (31fca3adb 2026-06-26)
 - fjall 3.16 + lsm-tree 3.16

marvin-j97:
`L0 needs to be disjoint` implies that your application wrote some keys that makes the key range of a newly flushed table overlap with a previous table. "FIFO" is kind of a misnomer which I adapted from RocksDB, it makes much more sense for it to be called `Monotonic` or something. It requires all keys to be in strict monotonic order (keys x, y, z with T(x) < T(y) < T(z) must be x < y < z, or the other way around).
When the FIFO compaction strategy detects an overlap in L0, it panics, which poisons some lock (forever), because there is no differentiation between recoverable (this) and unrecoverable (IO) errors right now. Really it should probably just log a warning and then do nothing...

[...]

PR #314, opened 2026-07-31
fix(fifo): use Lmax as canonical level

### LINKED ISSUE #312: Adjust FIFO compaction to better handle major compaction
Even though there is no reason to really major compact a FIFO-tree, it should still not screw it up.
FIFO compaction puts all tables into L0, but major compaction moves them all into Lmax.
So FIFO compaction should probably always operate on Lmax.

Would allow to repair a situation like #310

hhe48203-ctrl:
## What

- Treat Lmax as the canonical FIFO level.
- Move disjoint L0 tables into Lmax without rewriting them.
- Merge L0 with overlapping Lmax tables when a move would violate disjointness.
- Apply FIFO TTL and size drops to Lmax, including after major compaction.

## Why

Major compaction places all tables in Lmax, but FIFO only inspected L0. Once L0 was empty, FIFO became a no-op and could no longer enforce its TTL or size limit.

This keeps L0 as an ingestion level and makes FIFO compaction a two-step process when needed: first move or merge L0 into Lmax, then evaluate drops from Lmax on the next compaction choice.

The separate L0 overlap repair discussed in #310 and #311 remains out of scope; the existing disjoint-L0 invariant is preserved.

Fixes #312.

## Checks

- cargo fmt --all -- --check
- cargo clippy
- cargo test --all-features
- cargo test --doc --features lz4

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved FIFO compaction so level-0 tables move and merge correctly with final-level tables.
  - Size- and TTL-based cleanup now reliably removes eligible tables.
  - Improved handling of busy or empty levels to avoid unnecessary compaction.
  - Repeated compaction now behaves consistently and helps maintain blob-file counts within configured limits.

- **Tests**
  - Expanded coverage for table movement, overlapping merges, repeated compactions, and FIFO retention limits.

### DIFF
--- a/src/compaction/fifo.rs
+++ b/src/compaction/fifo.rs
@@ -191,7 +191,9 @@ impl CompactionStrategy for Strategy {
 mod tests {
     use super::Strategy;
     use crate::{AbstractTree, Config, KvSeparationOptions, SequenceNumberCounter};
-    use std::sync::Arc;
+    use std::sync::{Arc, Mutex};
+
+    static TIME_OVERRIDE_LOCK: Mutex<()> = Mutex::new(());
 
     #[test]
     fn fifo_empty_levels() -> crate::Result<()> {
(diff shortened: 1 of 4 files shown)


src/compaction/fifo.rs
```diff
@@ -71,23 +71,60 @@ impl CompactionStrategy for Strategy {
...
+                .map(Table::id)
+                .collect();
+
+            table_ids.extend(&overlapping_table_ids);
+
+            let input = CompactionInput {
+                table_ids,
+                dest_level: last_level_idx,
+                canonical_level: last_level_idx,
+                target_size: u64::MAX,
```
marvin-j97: We don't have a FIFO table target size, but a sensible default size around 128-256 MB would be better than `u64::MAX`, because that means a single possibly giant table, which makes TTL harder, and can make compactions really expensive if we actually end up having overlaps.

marvin-j97 (changes requested):

PR #319, opened 2026-08-09
test: fuzz optimize_runs visibility

### LINKED ISSUE #318: Fuzz test optimize_runs
`optimize_runs` is an important function that needs to be thoroughly tested (more than it currently is), and regressions would be awful. A fuzz test should be pretty easy to implement as the function is generic, so we can easily build test data without creating a real database.
As fuzz input, there are generally two operations: `insert(K, Seqno)` into an in-memory buffer (B-tree),  and `flush()`, which writes the buffer to a faux "table", which holds a key range. After each flush, all tables should run through `optimize_runs`, and the last written version of any key should always be visible.

fzlzjerry:
Fixes #318

## Summary

- add an AFL target that models inserts and flushes with in-memory faux tables
- compare optimized point visibility with a `BTreeMap` last-write model after every flush
- assert table preservation and per-run range disjointness
- expose a narrow, doc-hidden adapter around the production optimizer
- include a deterministic transitive-overlap regression seed

## Validation

- `cargo test --manifest-path fuzz/optimize_runs/Cargo.toml`
- `cargo afl build --manifest-path fuzz/optimize_runs/Cargo.toml`
- 10-second AFL smoke run: 0 crashes, 0 timeouts
- `cargo test --all-features`
- `cargo test --doc --features lz4`
- `cargo clippy`
- `cargo clippy --manifest-path fuzz/optimize_runs/Cargo.toml --no-deps -- -D warnings`
- `cargo fmt --all -- --check`
- `cargo fmt --manifest-path fuzz/optimize_runs/Cargo.toml -- --check`
- `git diff --check`

## Summary by CodeRabbit

* **Testing**
  * Added automated fuzz testing for data-run optimization, including inserts, flushes, overlapping ranges, ordering, table coverage, and value visibility.
  * Added regression coverage for complex transitive overlaps and newest-value selection.

* **Reliability**
  * Expanded validation of run coverage, ordering, overlap handling, and visible values.

* **Chores**
  * Added supporting infrastructure for storage reliability and optimization testing.

### DIFF
--- a/fuzz/optimize_runs/src/main.rs
+++ b/fuzz/optimize_runs/src/main.rs
@@ -92,7 +92,10 @@ fn run_operations(operations: impl IntoIterator<Item = Operation>) {
 fn main() {
     fuzz!(|data: &[u8]| {
         let mut unstructured = Unstructured::new(data);
-        let Ok(operations) = Vec::<Operation>::arbitrary(&mut unstructured) else {
+        let Ok(operations) = unstructured.arbitrary_iter::<Operation>() else {
+            return;
+        };
+        let Ok(operations) = operations.take(256).collect::<arbitrary::Result<Vec<_>>>() else {
             return;
         };
 
--- a/fuzz/optimize_runs/src/main.rs
+++ b/fuzz/optimize_runs/src/main.rs
@@ -1,0 +1,145 @@
+#[macro_use]
+extern crate afl;
+
+use arbitrary::{Arbitrary, Unstructured};
+use lsm_tree::{KeyRange, fuzzing::optimize_runs};
+use std::collections::BTreeMap;
+
+#[derive(Arbitrary, Debug)]
+enum Operation {
+    Insert { key: u8, seqno: u64 },
+    Flush,
+}
+
+type Table = BTreeMap<u8, u64>;
+type Runs = Vec<Vec<(usize, KeyRange)>>;
+
+fn verify(runs: &Runs, tables: &[Table], expected: &BTreeMap<u8, u64>) {
+    let mut table_ids = runs
+        .iter()
+        .flat_map(|run| run.iter().map(|(id, _)| *id))
+        .collect::<Vec<_>>();
+    table_ids.sort_unstable();
+    assert_eq!(table_ids, (0..tables.len()).collect::<Vec<_>>());
+
+    for run in runs {
+        for (index, (_, range)) in run.iter().enumerate() {
+            for (_, other) in run.iter().skip(index + 1) {
+                assert!(
+                    !range.overlaps_with_key_range(other),
+                    "optimized run contains overlapping tables: {run:?}"
+                );
+            }
+        }
+    }
+
+    for (&key, &expected_seqno) in expected {
+        let actual = runs.iter().find_map(|run| {
+            run.iter()
+                .find(|(_, range)| range.contains_key(&[key]))
+                .and_then(|(id, _)| tables[*id].get(&key).copied())
+        });
+
+        assert_eq!(
+            actual,
+            Some(expected_seqno),
+            "wrong visible version for key {key}: runs={runs:?}"
+        );
+    }
+}
+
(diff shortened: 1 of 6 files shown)


fuzz/optimize_runs/src/main.rs
```diff
@@ -0,0 +1,145 @@
...
+}
+
+fn main() {
+    fuzz!(|data: &[u8]| {
+        let mut unstructured = Unstructured::new(data);
+        let Ok(operations) = Vec::<Operation>::arbitrary(&mut unstructured) else {
+            return;
+        };
+
+        run_operations(operations);
```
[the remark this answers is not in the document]
fzlzjerry: Fixed in `7f7dbc2`. The harness now uses `arbitrary_iter`, applies the 256-operation limit before collecting, and therefore decodes and allocates at most 256 operations. The fallible collection still completes before `run_operations` is called, so malformed input does not execute a partial prefix.

I also added a unit test for the production optimizer adapter paths that were uncovered in the patch report. Validation passed with the fuzz-target test, 439 nextest tests, 23 documentation tests, formatting, and the repository's clippy command.

src/version/mod.rs
```diff
@@ -9,6 +9,73 @@ pub mod recovery;
 pub mod run;
 mod super_version;
 
+#[doc(hidden)]
```
marvin-j97: Can this be defined in the fuzz harness instead?
The necessary exports (just as Ranged) can be exported as hidden from lib.

fzlzjerry: Done in `eebf6bc`. The adapter, `FuzzTable`, conversions, and adapter test now live in the fuzz harness. The library only exposes the three harness requirements—`optimize_runs`, `Ranged`, and `Run`—as doc-hidden items. Both the full crate tests and the fuzz-harness tests pass.

fzlzjerry:
Addressed the [out-of-diff sorted-range finding](https://github.com/fjall-rs/lsm-tree/pull/319#pullrequestreview-4905544816) in `ad2a73ca`.

`verify` now requires each adjacent pair to satisfy `previous.max() < next.min()`, matching the sorted/disjoint precondition of `Run::get_for_key`. The existing table-preservation, pairwise-overlap, and visibility checks remain in place. No library API changes.

The regression first accepts correctly ordered disjoint ranges, then reverses the same run and requires rejection. It failed against the previous validator because linear visibility lookup still accepted the reversed run, and passes with the new invariant.

Validation on macOS arm64, Rust/Cargo 1.97.0:
- `cargo +stable test --manifest-path fuzz/optimize_runs/Cargo.toml`: all 3 harness tests passed.
- `cargo +stable clippy --manifest-path fuzz/optimize_runs/Cargo.toml --all-targets -- -D warnings`: passed; the dependency library still emits its two existing dead-code warnings.
- Harness formatting checked.

fzlzjerry:
Addressed the [production lookup boundary suggestion](https://github.com/fjall-rs/lsm-tree/pull/319#pullrequestreview-5189506124) in `bff22a4`.

The harness now keeps the optimizer’s real Run<FuzzTable> values and resolves visibility through Run::get_for_key. This removes the tuple conversion and range cloning rather than constructing new wrappers for every checked key. Table preservation and the sorted/disjoint invariant remain; the redundant all-pairs overlap scan is gone because strict adjacent disjointness already implies it.

The new boundary/gap scenario checks table maxima, disjoint tables, and falling through a newer table’s key gap to older data. I also replaced the adapter-only assertion with this behavioral coverage and made the reversed-run regression require the ordering assertion specifically.

Mutation check in an isolated worktree: changing the production partition boundary from max() < key to max() <= key leaves the old harness’s 3 tests green; with the new harness it fails both visibility tests at missing maximum keys. The real branch does not contain that mutation or any library changes.

Final harness tests: 3 passed. Clippy and formatting passed on Rust/Cargo 1.97.0; the dependency library still emits its two existing dead-code warnings.

PR #320, opened 2026-08-29
fix: decline compaction results whose input tables were removed by clear()

samvallad33:
Fixes the crash reported in fjall-rs/fjall#287.

Compaction runs in two steps. First it reads the version and picks tables to move. Then it takes a write lock and commits. Nothing holds the version steady in between.

Tree::clear takes that same write lock and installs an empty version, so it can land right in the gap. The commit then asserts every input table still exists, finds none, and panics.

The poisoning is the real damage. That assert fires while the worker holds both the compaction lock and the version write lock, so both are poisoned. One lost race kills the keyspace permanently.

This change makes the commit step check that its planned tables are still present, and decline the compaction if they are not. A clear that lands mid-compaction now discards that compaction instead of taking the process down.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented compaction from committing stale results when table versions change concurrently.
  * Improved safety when clearing data during compaction, avoiding invalid table merges or moves.
  * Compaction now safely declines outdated operations instead of applying results to replaced data.

* **Tests**
  * Added coverage for race conditions involving concurrent clearing and compaction.

### DIFF
--- a/src/compaction/worker.rs
+++ b/src/compaction/worker.rs
@@ -203,6 +203,28 @@ fn move_tables(
 
     let table_ids = payload.table_ids.iter().copied().collect::<Vec<_>>();
 
+    // NOTE: Check for race condition
+    //
+    // The compaction choice was made under a *read* lock which we dropped before taking
+    // this write lock, so the version may have been swapped out in between (`clear()` for
+    // example installs a brand new, empty version). `Version::with_moved` asserts that every
+    // input table is still present, so committing a stale move would panic while holding both
+    // the compaction state mutex and the version history write lock, poisoning them.
+    //
+    // Fixes: https://github.com/fjall-rs/fjall/issues/287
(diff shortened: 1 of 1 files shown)


src/compaction/worker.rs
```diff
@@ -961,3 +1012,140 @@ mod tests {
...
+
+    impl CompactionFilter for ClearMidCompactionFilter {
+        fn filter_item(
+            &mut self,
+            _: ItemAccessor<'_>,
+            _: &FilterContext,
+        ) -> crate::Result<Verdict> {
+            if let Some(tree) = self.0.lock().unwrap().take() {
+                tree.clear()?;
+            }
```
marvin-j97, line 1086: This is incredibly disgusting but also kinda clever

marvin-j97 (approved):

